PR #30, opened 2024-06-20
Fix build and release new patch

woksin:
## Summary

Fixes a problem in the build pipeline that hindered the release of the previous 2 nuget packages


woksin:
This was fixed by #29

PR #34, opened 2024-07-01
Adding client concerns

einari:
### Added

- Adding support for ignoring validation in the Command and Query pipelines by adding a new attribute in front of either a controller class or an action method. `[IgnoreValidation]`.
- Adding support for applying paging from the client without having to implement paging in the backend by supporting well-known query types such as `IQueryable`. By returning a queryable in the Query controller action, it will add `.Skip().Take()` statements correctly.

### Fixed

- Fixes a bug in the proxy generator where it treated query arguments that were concept types as complex types and flattened them to be called `value`. Obviously not the correct expected name, and could cause conflicts when having multiple arguments.


woksin:
I fixed the issue with paging, it was due to the handling of state using useState that screwed up a lot of the things. I made the useQuery code a lot simpler and had to remove the private _ fields from the QueryFor type because it caused problems due to React's useState shallow copying

woksin:
Sorting is not working at all, so I just skipped that for now as I have no idea how to fix it

PR #43, opened 2024-07-19
Add ability to set init store to MVVM

maspasov:
### Added

MVVM now has a prop for the init store. We can define global store and observe into every component for some changes in that store. 
I fixed an error in the Identity provider in case of no valid response from that API.

### Changed

- MVVM Provider
- Identity Provider

### Fixed

image

Example how we can set init store from some App:

image


einari:
A couple of things:

1. Could you fix the description to be more like we have for all of our releases. The pictures don't really explain what has been changed / added / fixed in the release. For inspiration, have a look [here](https://github.com/Cratis/ApplicationModel/releases).
2. Please set the correct labels for what type of semantic version change this represents. 
image
3. Could you explain what this change does? Since we already have the `Mobx` options on the `MVVMProps` and `initStore`is not a concept in application model itself, I'm struggling to see if the introduction of this property will affect anything?

maspasov:
> A couple of things:
> 
> 1. Could you fix the description to be more like we have for all of our releases. The pictures don't really explain what has been changed / added / fixed in the release. For inspiration, have a look [here](https://github.com/Cratis/ApplicationModel/releases).
> 2. Please set the correct labels for what type of semantic version change this represents.
>    image
> 3. Could you explain what this change does? Since we already have the `Mobx` options on the `MVVMProps` and `initStore`is not a concept in application model itself, I'm struggling to see if the introduction of this property will affect anything?

maspasov:
> A couple of things:
> 
> 1. Could you fix the description to be more like we have for all of our releases. The pictures don't really explain what has been changed / added / fixed in the release. For inspiration, have a look [here](https://github.com/Cratis/ApplicationModel/releases).
> 2. Please set the correct labels for what type of semantic version change this represents.
>    image
> 3. Could you explain what this change does? Since we already have the `Mobx` options on the `MVVMProps` and `initStore`is not a concept in application model itself, I'm struggling to see if the introduction of this property will affect anything?

I have added initStore prop because it will be used for MVVM provider, mobx prop is a specific property for configuring Mobx library. There is a perfect example of what we can do with the store provider which is MVVM in our case. https://youtu.be/pnhIJA64ByY?t=261

maspasov:
We decided not to add a store to the MVVM provider. So for now I will close that PR and add fix regarding IdentityProvider as a separate PR.

maspasov:
We decided not to add a store to the MVVM provider. So for now I will close that PR and add fix regarding IdentityProvider as a separate PR.

PR #77, opened 2024-08-17
Feature: Serializing Dates and time MongoDb

### LINKED ISSUE #76: Add MongoDB serializer support for concepts of DateTime, DateTimeOffset, DateOnly, TimeOnly
We have a `ConceptSerializer` it needs to do the right thing when both serializing and deserializing concepts of DateTime, DateTimeOffset, DateOnly and TimeOnly.

DateTimeOffset we want to be consistent with and serialize using the `DateTimeOffsetSupportingBsonDateTimeSerializer`

smellilac:
### Fixed

- Fixing support for `DateTime`, `DateOnly`, `TimeOnly` and `DateTimeOffset` serialization of `ConceptAs<>` types of these types (#76)

### DIFF
--- a/Source/DotNET/MongoDB/ConceptSerializer.cs
+++ b/Source/DotNET/MongoDB/ConceptSerializer.cs
@@ -5,7 +5,6 @@ using Cratis.Concepts;
 using MongoDB.Bson;
 using MongoDB.Bson.IO;
 using MongoDB.Bson.Serialization;
-using System.Globalization;
 
 namespace Cratis.Applications.MongoDB;
 
@@ -138,13 +137,13 @@ public class ConceptSerializer<T> : IBsonSerializer<T>
         }
         else if (underlyingValueType == typeof(DateOnly))
         {
-            var dateOnly = (DateOnly)underlyingValue;
-            bsonWriter.WriteString(dateOnly.ToString("yyyy-MM-dd"));
+            var serializer = new DateOnlySerializer();
+            serializer.Serialize(context, args, (DateOnly)underlyingValue);
         }
         else if (underlyingValueType == typeof(TimeOnly))
         {
-            var timeOnly = (TimeOnly)underlyingValue;
-            bsonWriter.WriteString(timeOnly.ToString("HH:mm:ss"));
+            var serializer = new TimeOnlySerializer();
+            serializer.Serialize(context, args, (TimeOnly)underlyingValue);
         }
     }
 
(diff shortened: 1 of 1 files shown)


einari:
Sorry for getting back so late. This is brilliant. Issue #76 failed to mention that we have a `TimeOnlySerializer` and a `DateOnlySerializer` that we could be forwarding to directly, as you've done with `DateTimeOffsetSupportingBsonDateTimeSerializer`. Great if you could fix that :)

smellilac:
> Sorry for getting back so late. Thanks for your contribution, this is brilliant. Issue #76 failed to mention that we have a `TimeOnlySerializer` and a `DateOnlySerializer` that we could be forwarding to directly, as you've done with `DateTimeOffsetSupportingBsonDateTimeSerializer`. Great if you could fix that :)
> 
> Also, we have a few static code analysis rules and run with "TreatWarningsAsErrors". A tip is to run `dotnet build -c release` before pushing.

Thank you very much for the kind words and advice!

I have a couple of questions about this code:

```
else if (underlyingValueType == typeof(DateTimeOffset))
{
var serializer = new DateTimeOffsetSupportingBsonDateTimeSerializer();
serializer.Serialize(context, args, (DateTimeOffset)underlyingValue);
}
else if (underlyingValueType == typeof(DateOnly))
{
var serializer = new DateOnlySerializer();
serializer.Serialize(context, args, (DateOnly)underlyingValue);
}
else if (underlyingValueType == typeof(TimeOnly))
{
var serializer = new TimeOnlySerializer();
serializer.Serialize(context, args, (TimeOnly)underlyingValue);
}
```

1) underlyingValue is an object? (reference type), but we cast it to a struct (value type), are you okay with that?

2) Should we do explicit casts, maybe it's better to use ``as``?

einari:
The explicit cast is fine as you have it. Since they’re not reference types, `as` won’t work. 

If you’re good, we can pull this in and merge.
Btw: I have fixed the summary, since that is part of the changelog on the GitHub releases.

einari (approved):

smellilac:
> The explicit cast is fine as you have it. Since they’re not reference types, `as` won’t work.
> 
> If you’re good, we can pull this in and merge. Btw: I have fixed the summary, since that is part of the changelog on the GitHub releases.

Thank you for the clarification and for fixing the summary. 

Everything looks good to me now. I appreciate the review and the work done.

einari:
Great. And likewise. Thanks again for the contribution.

PR #333, opened 2024-10-22
Append header to Commands

panayot-zhi:
### Changed

- use private variable for microservice name instead of Globals.microservice


woksin (approved):
LGTM

Issue #2152, opened 2026-04-18
Read Model interception


einari:
We need to be able to intercept read models before they're served to a client. 

This needs to be consistently done for regular queries, both regular ones and for observable ones.

The purpose of this is to enable cross cutting operations on read models, such as decryption or other things. 

We have our query pipeline where this fits. 

We are going to need specs for both with and without AspNet - controller based and model bound.

PR #2160, opened 2026-04-21
Add read model interception for queries and observable streaming

woksin:
## Added

- `IInterceptReadModel<TReadModel>` interface for implementing cross-cutting read model interceptors (decryption, masking, etc.) (#2152)
- `IReadModelInterceptors` applied consistently across model-bound queries, controller-based queries (`QueryActionFilter`), and all observable streaming paths (`ClientObservable`, `ClientObservableSSE`, `ClientEnumerableObservable`, `ClientEnumerableObservableSSE`) (#2152)
- Interceptors are discovered via `ITypes` at startup and resolved from `IServiceProvider` per request, supporting scoped and transient lifetimes (#2152)


woksin:
@copilot fix test issues

einari:
@copilot it should use the ITypes from Fundamentals for discovery of interceptor. See other implementations here. That means one does not haven't register interceptors in the ServiceCollection. Update the docs to remove this requirement.

einari:
@copilot Lets make the IReadModelInterceptors Intercept method take an IEnumerable<object> instead and let it do the looping. For those places where it's a single read model we just pass [readModel]

woksin:
@einari this looks good now

einari:
> @einari this looks good now

actually, I just noticed something. There is a bad assumption it can mutate in place, record types which we typically use, can't do that. So I'm going to issue another comment co CoPilot

einari:
@copilot The assumption of in place editing is wrong. When using record types for instance, these are immutable. So the interceptors need to return the changed read model instead.

einari:
@copilot In the ReadModelInterceptors, we should parallelize every call to the intercept method for each read model instance.

woksin:
@copilot Interceptor can probably return ValueTask instead of Task. That's a performance gain. After that make sure that the documentation is up to date and good

einari:
@copilot merge main (rebase) and fix any conflicts

einari:
@copilot Fix the broken specs

PR #2192, opened 2026-05-04
Fix ProxyGenerator type ordering for property type annotation forward references

einari:
## Fixed

- `ReferenceError: Cannot access '...' before initialization` in combined generated TypeScript files when a property uses `@field(String)` but its type annotation references another generated class (e.g. `all!: FromEveryDefinition`). TypeScript's `emitDecoratorMetadata` emits `__metadata("design:type", T)` based on the declared type annotation, not the `@field` argument — so the referenced class must appear earlier in the file.


einari (commented):
@copilot See the code review and implement as suggested; https://github.com/Cratis/Arc/pull/2192#pullrequestreview-4221872096

PR #2273, opened 2026-06-04
Add @cratis/eslint-plugin-arc

woksin:
## Added
- `@cratis/eslint-plugin-arc`: ESLint rules for projects that consume Cratis Arc, designed to compose on top of `@cratis/eslint-config`.
  - **`skip-generated-proxies`** (processor) — skips Arc-generated proxy files wholesale. They carry a `// @generated by Cratis` header (and a `**DO NOT EDIT**` banner), cannot be edited, and are regenerated by the build, so any lint finding on them is un-actionable. Keyed on the header because generated proxies sit intermixed with hand-written `.ts` under the same folders, with no path distinction. The pass-through preserves filenames (import resolution is unaffected) and keeps `--fix` working on hand-written files.
  - **`no-hooks-in-view-model`** (rule) — forbids React hook calls inside MVVM view models (classes named `*ViewModel`). A view model must be a plain, React-free class; inject Cratis abstractions instead of calling hooks. Only bare-identifier hook calls are flagged, so view-model methods that merely start with `use` are not false positives. The class suffix and hook pattern are configurable.

## Notes
- Named per the ESLint scoped-plugin convention (`@scope/eslint-plugin-*`). The plugin namespace is `@cratis/arc`, so rule IDs read `@cratis/arc/no-hooks-in-view-model`.
- Ships as raw ESM (no build step), declares only `eslint` as a peer dependency, and exposes `meta.version` plus per-rule docs URLs. It owns these rules because Arc owns the proxy generator and `@cratis/arc.react.mvvm`.


woksin:
## ⚠️ One-time first-publish bootstrap (before this releases cleanly)

This PR adds a **brand-new npm package** — `@cratis/eslint-plugin-arc`. The Publish workflow uses npm **trusted publishing (OIDC)** (it strips the auth token), which requires a trusted-publisher relationship that must already exist on the package. That can't be configured before the package exists, so the package's **first** publish over OIDC will fail.

One-time bootstrap (do before merge, or after — the companion resilience commit means an un-bootstrapped package no longer blocks the others):

- [ ] **Initial token-based publish** to create the package, authenticated to npm with publish rights to `@cratis`:
  ```sh
  cd Source/JavaScript/Arc.ESLint
  npm publish --access public
  ```
  (or run with a granular `NODE_AUTH_TOKEN`). `publishConfig.access: public` is already set.
- [ ] **Configure the trusted publisher**: npmjs.com → `@cratis/eslint-plugin-arc` → *Settings → Trusted Publisher* → repository `Cratis/Arc`, workflow `.github/workflows/publish.yml`.
- [ ] **Verify** the next Publish run publishes over OIDC with no token.

**Heads-up — co-versioning:** `publish-version` sets every workspace to the repo release version, so on the first automated release this package jumps to ~`20.x` (matching `@cratis/arc`), not `0.x`/`1.0.0`. Starting it independently would need a release-pipeline change.

woksin:
@einari — two one-time steps before merging this. This is **1 of 3 coordinated PRs**: Fundamentals Cratis/Fundamentals#1072, Arc #2273, Components Cratis/Components#95.

### 1. Bootstrap `@cratis/eslint-plugin-arc` on npm (one-time)
The Publish workflow uses OIDC trusted publishing, which can't create a package that doesn't exist yet — so the first publish has to be done by hand. From a checkout of this branch, authenticated to npm with publish rights to `@cratis`:
```sh
cd Source/JavaScript/Arc.ESLint
npm publish --access public
```
Then on npmjs.com → **@cratis/eslint-plugin-arc** → *Settings → Trusted Publisher*: repository `Cratis/Arc`, workflow `.github/workflows/publish.yml`. (`publishConfig.access: public` is already set; the throwaway `0.0.0` gets superseded by the real version on the first automated release.)

### 2. Merge with the `minor` label (already applied)
Merging triggers the Publish workflow, which co-versions and releases every workspace over OIDC — including the now-bootstrapped package.

**Order:** bootstrap → merge for a fully-green first release. Merging before bootstrapping is safe too — the resilience commit means this new package failing to publish no longer blocks `@cratis/arc` or the other workspaces; it just won't publish until bootstrapped.

**Versioning note:** co-versioning makes this plugin debut at ~`20.x` (the repo version), not `1.0.0`. Say the word if you want it versioned independently and we'll adjust the pipeline first.

PR #2275, opened 2026-06-05
Fix observable queries breaking under read model interception

woksin:
# Summary

The Chronicle read model interceptor (PII release) is registered for every read model, so `QueryPipeline.ApplyInterceptors` ran it against observable (`ISubject<T>`) and async-enumerable query results. It handed the streaming wrapper to `IInterceptReadModel<TReadModel>.Intercept(TReadModel)`, throwing `Object of type 'LifetimeAwareSubject<…>' cannot be converted to type '<ReadModel>'` and breaking every observable model-bound query served over the multiplexed SSE/WebSocket hub. Streaming interception now lives behind a single helper shared by all streaming transports, so compliance/PII release runs per emission — including for collection queries, which previously skipped it.

## Fixed

- Observable (`ISubject<T>`) and `IAsyncEnumerable<T>` model-bound query results no longer crash under read model interception. `QueryPipeline.ApplyInterceptors` now leaves streaming results untouched for the streaming transport to intercept per emission.
- Collection observable queries (`ISubject<IEnumerable<TReadModel>>`) now have compliance/PII release applied to each item. The streaming transports previously keyed interception on the enumerable type, found no matching interceptor, and silently served undecrypted values.

## Changed

- All streaming transports (`ObservableQueryDemultiplexer`, `ClientObservable`, `ClientObservableSSE`, `ClientEnumerableObservable`, `ClientEnumerableObservableSSE`) now share a single `IReadModelInterceptors.InterceptEmission` helper that intercepts each emission, unwrapping collection emissions to the read model element type.

### DIFF
--- a/Source/DotNET/Arc.Core.Specs/Queries/for_ObservableQueryDemultiplexer/given/an_observable_query_demultiplexer.cs
+++ b/Source/DotNET/Arc.Core.Specs/Queries/for_ObservableQueryDemultiplexer/given/an_observable_query_demultiplexer.cs
@@ -14,6 +14,8 @@ public class an_observable_query_demultiplexer : Specification
     protected IQueryContextManager _queryContextManager;
     protected IHttpRequestContextAccessor _httpRequestContextAccessor;
     protected IHostApplicationLifetime _hostApplicationLifetime;
+    protected IReadModelInterceptors _readModelInterceptors;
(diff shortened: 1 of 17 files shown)


einari:
@woksin Looks good to me. Great if you do a second review of it, if you haven't already before merging.

PR #2310, opened 2026-06-21
Fix Cratis meta-package and Chronicle consumption for downstream apps

woksin:
Makes Arc consumable again from the `Cratis` meta package alongside Arc.Chronicle. Arc 20.44.0 broke real consumers at build and runtime; this restores a clean restore-build-run path against released Chronicle.

## Changed

- Raised the minimum required `Cratis.Chronicle` to `15.35.0`.

## Fixed

- Building a project that references the `Cratis` meta package together with a package that transitively brings `Cratis.Arc.Core` (such as `Cratis.Arc.MongoDB`) no longer fails with `CS0101 'Cratis.Arc.Generated.GeneratedMarker'` caused by the Arc source generator being loaded and run twice.
- Arc no longer throws `MissingMethodException` at runtime against released Chronicle. The Chronicle floor now matches the `AppendMany`/`Append` overloads Arc actually calls, so an incompatible Chronicle version fails `dotnet restore` instead of crashing at runtime.
- The `ARCCHR0002` analyzer no longer reports an ambiguous command event source id when the command's `Handle` supplies the event source explicitly through a tuple return — including new-stream-create commands that return a generated id deriving from `EventSourceId<T>`.

### DIFF
--- a/Directory.Packages.NET8-9.props
+++ b/Directory.Packages.NET8-9.props
@@ -6,8 +6,8 @@
         <PackageVersion Update="Microsoft.EntityFrameworkCore.Sqlite" Version="9.0.16" />
         <PackageVersion Update="Microsoft.EntityFrameworkCore.SqlServer" Version="9.0.16" />
         <PackageVersion Update="Npgsql.EntityFrameworkCore.PostgreSQL" Version="9.0.4" />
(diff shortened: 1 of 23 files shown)


woksin:
Superseded by #2312, which consolidates this PR and #2311 into a single PR (and adds the per-TFM dependency baseline fix). Closing in favor of #2312.

PR #2311, opened 2026-06-21
Expose and document the MVVM observer boundary

woksin:
Makes the sanctioned MobX observer boundary in Cratis Arc React MVVM explicit, documented, and guarded, so consumer apps stop reaching through the transitive MobX dependency or blanket-wrapping components.

## Added

- Exported `observer` from `@cratis/arc.react.mvvm` so a child component that reads view model observables directly can be given its own observer boundary, with guidance in the React MVVM documentation.
- Added a `no-direct-mobx-react-import` ESLint rule that flags direct imports from `mobx-react` / `mobx-react-lite` and points consumers to `@cratis/arc.react.mvvm` instead.

## Changed

- The recommended Cratis Arc ESLint config now reports direct `mobx-react` / `mobx-react-lite` imports as errors.


woksin:
Superseded by #2312, which consolidates this PR and #2310 into a single PR. Closing in favor of #2312.

PR #2328, opened 2026-07-08
Add Arc analyzers for command Handle, [Roles], and exception hygiene

woksin:
## Added

- Analyzer (ARC0010) that flags a `[Command]` `Handle()` returning `Task`/`Task<T>` without awaiting anything — including the `Task.FromResult`/`Task.CompletedTask` wrapper shape the compiler's own CS1998 does not catch — with a code fix that unwraps it to the synchronous return shape the command pipeline already accepts.
- Analyzer (ARC0011) that flags a `[Roles(...)]` string-literal argument, with a code fix that rewrites it to `nameof(<EnumType>.<Member>)` so a role rename becomes a compile error instead of a silent authorization failure.
- Analyzer (ARC0012) that flags built-in exception types (`Exception`, `InvalidOperationException`, `ArgumentException`, and the like) thrown from Arc artifacts — command `Handle()` methods, `CommandValidator<T>`/`ConceptValidator<T>` validators, and reactor handlers — steering domain failures toward domain-named exception types.


woksin:
Folded into #2329 — the analyzer commits were cherry-picked onto that branch and the release notes moved to its description. Closing this standalone PR.

PR #2355, opened 2026-07-18
Add ARC0013: flag a validator dereferencing a possibly-null concept

woksin:
## Added

- New analyzer **ARC0013** — warns when a validator rule dereferences a member of a possibly-null concept (for example `RuleFor(c => c.Order.Value)`). Because model-bound input can be deserialized with a null concept, the rule throws at validation time instead of producing a validation error; the analyzer flags it in the editor and points to the fixes (validate the concept when required, or guard the rule with `.When(...)`).


woksin:
Consolidated into #2354 — all resilience and analyzer changes now ship in a single PR, as requested.

Issue #2351, opened 2026-07-15
Header tenant resolution silently yields no tenant: IHttpRequestContextAccessor is not populated for the whole request


einari:
## Summary

In a production ASP.NET Core application (Cratis Studio), header-based tenant resolution silently never resolved a tenant — every request executed against the Default Chronicle namespace and the un-suffixed MongoDB database, across organizations. Replacing `HeaderTenantIdResolver` with an application-owned `ITenantIdResolver` backed by ASP.NET's `IHttpContextAccessor` fixed it immediately (verified with MongoDB profiling: the identity lookup moved from `Studio.…` to `Studio+cratis.…` and authorization started resolving the correct per-tenant role).

## Root cause analysis

`HeaderTenantIdResolver` (and the other resolvers) read the request through `IHttpRequestContextAccessor`. That accessor's `Current` is only assigned inside `AspNetCoreEndpointMapper.MapGet/MapPost` request delegates, `WellKnownRoutesMiddleware`, and the observable-query demultiplexer. Anything that triggers tenant resolution earlier — or on a different async path — sees `Current == null` and resolves no tenant:

- `TenantIdAccessor` caches the result in a `static AsyncLocal<TenantId>`; once a resolution happens with no request context, `NotSet` sticks for the rest of that flow.
- The scoped `IMongoDatabase` resolves its database name once per scope via `IMongoDatabaseNameResolver` → `ITenantIdAccessor`. When identity resolution (e.g. an `IProvideIdentityDetails` implementation reading tenant-scoped collections) creates that scoped database before/outside the endpoint delegate's assignment of `Current`, the Default database is captured for the whole request.

Two adjacent issues spotted while diagnosing:

1. `AspNetCoreHttpRequestContext.Headers` builds its dictionary with `ToDictionary(kvp => kvp.Key, …)` — an ordinal, case-sensitive lookup. HTTP/2 lowercases header names on the wire, so a configured header like `Tenant-ID` can never match an HTTP/2 request. Header lookups should be case-insensitive (`StringComparer.OrdinalIgnoreCase`).
2. The XML docs on `UseHeaderTenancy`/`UseSubdomainTenancy` say the default header is `'X-Tenant-ID'`, but `Constants.DefaultTenantIdHeader` is `x-cratis-tenant-id`. (The `Constants` class doc summary also says "constants related to correlation id".)

## Suggested fix

[...]

Issue #2303, opened 2026-06-17
Support injecting read models from non-Chronicle providers into the command pipeline


woksin:
## Background

In Arc, a read model is injectable into command-scoped code (a `CommandValidator<>`, `Provide()`, or `Handle()`) only if some provider registers a **command-scope resolver** for its type that loads the instance **by key** (the resolved event source id) from the appropriate store. Command-handler arguments are resolved from DI by type via `CommandHandlerArgumentResolver` → `ParameterDependencyResolver.Resolve`.

Today, that registration exists **only** in the Chronicle integration (`Cratis.Arc.Chronicle`): `AddReadModels` registers a scoped resolver (`ResolveReadModel`) that loads read models through `IReadModels.GetInstanceById(...)`. It registers read models that have a Chronicle backing artifact — `IProjectionFor<T>` projections, model-bound projections, and `IReducerFor<T>` reducers.

`[ReadModel]` is an **Arc** concept (`Cratis.Arc.Queries.ModelBound`), broader than Chronicle. It can be backed by other stores (for example Entity Framework Core). Those read models are **not** currently injectable into the command pipeline, because no provider other than Chronicle registers a by-key command-scope resolver for them. EF Core currently registers DbContexts only (`AddReadModelDbContexts…`), with no per-read-model command resolver.

This was surfaced while fixing reducer-backed read model injection — the Chronicle `AddReadModels` must restrict itself to read models Chronicle can resolve by key, precisely so it does not wrongly claim non-Chronicle `[ReadModel]` types and try to resolve them through `IReadModels`.

## Problem

A `[ReadModel]` that is resolvable by key but is **not** backed by a Chronicle projection/model-bound projection/reducer (e.g. an EF Core-backed read model) cannot be injected into a command's `CommandValidator<>`, `Provide()`, or `Handle()`. There is no general, provider-agnostic mechanism for a backing provider to contribute a by-key command-scope resolver.

## Proposed direction

Provide a provider-agnostic way for any read-model backing provider to register a command-scope, by-key resolver for the read model types it owns, so injection works uniformly regardless of where the read model is materialized:

- A shared abstraction (e.g. an `IReadModelResolver`/by-key loader) that the command pipeline uses, with per-provider implementations (Chronicle via `IReadModels`, EF Core via the read model `DbContext`, etc.).
- Each provider registers the read model types it can resolve, keyed off the command context's resolved event source id (reusing the existing id/key resolution strategies).
[...]

Issue #2204, opened 2026-05-13
AOT / NativeAOT / Trimming / Single-file compatibility


einari:
## Goal

Make Cratis Arc fully compatible with:

- **PublishAot** (NativeAOT compilation)
- **PublishTrimmed** (IL Linker / trimming)
- **PublishSingleFile** (single-file publish)

This is a prerequisite for consumer apps (MAUI, minimal APIs, cloud-native) that target AOT or trimmed deployment profiles.

Reference: Cratis/Fundamentals#1043 — generated type discovery with AOT-safe metadata via Roslyn source generators (the direction Arc should follow for type discovery).

> Important: We do not want any breaking changes, so this change should be fully transparent 

---

### 1. Remove Castle.DynamicProxy + Polly resilience layer

**Files:** `Source/DotNET/MongoDB/Resilience/` (entire folder) + `MongoDB/MongoDBClientFactory.cs`

Castle.DynamicProxy generates proxy types at runtime — fundamentally incompatible with AOT/trimming. The entire folder should be deleted:

- `MongoClientInterceptor.cs`
- `MongoDatabaseInterceptor.cs`
- `MongoCollectionInterceptor.cs`
- `MongoCollectionInterceptorForReturnValues.cs`
- `MongoClientInterceptorSelector.cs`
- `MongoDatabaseInterceptorSelector.cs`
- `MongoCollectionInterceptorSelector.cs`
- `MongoClientDisposeInterceptor.cs`
- `RetryingChangeStreamCursor.cs`
- `EmptyAsyncCursor.cs`
- `WellKnownErrorMessages.cs`

`MongoDBClientFactory.cs` should be simplified to create a plain `MongoClient` with `RetryReads=true` and `RetryWrites=true` in the connection string / settings (the MongoDB driver has had built-in retry since v2.6/2.9 and it is on by default).

---

### 2. Replace `DependencyContext` + `Assembly.Load` type discovery

**File:** `Arc.Core/GeneratedMetadataRegistration.cs`

`DependencyContext.Default` is `null` in single-file publish; `Assembly.Load(new AssemblyName(...))` does not work in AOT. Replace with a source-generated provider pattern following the approach in Fundamentals PR #1043.

---

### 3. Fix `MakeGenericMethod` / `MakeGenericType` + `Invoke` patterns

These patterns are not statically analysable and will be trimmed away or throw under NativeAOT. Each site needs to be restructured to use closed generics knowable at compile time, or use source-generated dispatch tables.

[...]

Issue #1601, opened 2024-09-27
Add support for creating mementos from an aggregate root


einari:
It should be possible to configure the aggregate for how often the memento should be generated and it should also be possible for the aggregate to manually create mementos.

The definition / shape of the memento should be defined by the aggregate root itself, typically by a type.

Depends on Cratis/Chronicle#1370

woksin:
Triaging this for actionability. The direction is understood — configurable memento frequency, manual/on-demand memento creation, and an aggregate-defined memento shape — but there's no concrete API yet, and it references a dependency on Cratis/Chronicle#1370. @einari a few clarifications would let this be picked up:

- **Dependency status:** is Cratis/Chronicle#1370 resolved/available, or does this remain blocked on it? If blocked, this should probably be labeled/parked accordingly.
- **Frequency configuration:** where and how is the frequency expressed — per aggregate root type via attribute, fluent registration, or a global default with per-type override? Every N events, on a time interval, or both?
- **Manual creation API:** what's the intended call shape for creating a memento on demand (e.g. a method on the aggregate root, or via a service)?
- **Memento shape:** how does an aggregate define what goes into its memento — a dedicated method/type, or derived from the current state automatically?
- **Acceptance criteria:** a minimal example aggregate showing frequency config + a manual memento + the defined shape would make this verifiable.

Happy to implement once the API shape and the Chronicle#1370 dependency are settled.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #2367, opened 2026-07-22
Fix tenant resolution and remove AOT-incompatible validation reflection

woksin:
## Fixed

- Header-based tenant resolution now applies for the whole request instead of only inside endpoint delegates, so tenant-scoped work that runs earlier no longer falls back to the default namespace and database; header lookup is now case-insensitive so a configured header still matches when HTTP/2 delivers it lowercased (#2351)

### DIFF
--- a/Source/DotNET/Arc.Core/Tenancy/Constants.cs
+++ b/Source/DotNET/Arc.Core/Tenancy/Constants.cs
@@ -4,17 +4,17 @@
 namespace Cratis.Arc.Tenancy;
 
 /// <summary>
-/// Holds constants related to correlation id.
+/// Holds constants related to tenant id.
 /// </summary>
 public static class Constants
 {
     /// <summary>
-    /// Gets the header name for the correlation id.
+    /// Gets the default header name for the tenant id.
     /// </summary>
     public const string DefaultTenantIdHeader = "x-cratis-tenant-id";
 
     /// <summary>
-    /// Gets the item key for the correlation id.
+    /// Gets the item key for the tenant id.
     /// </summary>
     public const string TenantIdItemKey = "TenantId";
 }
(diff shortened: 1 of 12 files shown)


woksin:
### Reviewer notes (not release notes)

**Also in this PR (internal, no user-facing effect — deliberately kept out of the release-notes body):**
- Removed AOT-incompatible reflection from four validation sites — `BaseValidator` (`Expression.Compile()`), `RuleBuilderInitialExtensions` (private-member reflection), `DiscoverableValidators` (`MakeGenericType`), `ValidatorInvoker` (`MakeGenericType`). No behavior or public-API change; a self-contained slice of the AOT epic (#2204).

**Other open issues evaluated in this sweep, not shipped:**
- **#2303** (inject non-Chronicle read models into the command pipeline) — needs a core-pipeline design decision: the command "key" resolution (`GetEventSourceId` → `EventSourceId`) and `RegisteredReadModelTypes` live in Chronicle, and EF Core references only Arc.Core (not Chronicle, nor should it). Participating cleanly requires a provider-neutral command-key abstraction lifted into Arc.Core plus `RegisteredReadModelTypes` made additive across providers. Left for direction rather than guessed at.
- **#1493** (revise the command response value system) — a subsystem redesign needing a design decision.
- **#1601** (aggregate mementos) — blocked by the still-open Cratis/Chronicle#1370.
- **#64** (adhere to Microsoft library guidance) — broad and largely satisfied; the rest is packaging-sensitive and better scoped individually.

**Verification:** Arc.Core.Specs 1535 ✓, Arc.Specs 196 ✓, 5 new tenant specs ✓; Release builds of Arc and Chronicle both 0 warnings / 0 errors; all CI checks green.

woksin:
Folded into #2374 per request (kept as one PR).

PR #2368, opened 2026-07-23
Keep observable emissions safe across subscription teardown

woksin:
## Fixed

- An observable query no longer crashes the server when a client disconnects while an emission is in flight — emissions are dropped cleanly once the subscription is torn down, in either teardown order

### DIFF
--- a/Source/DotNET/Arc.Core.Specs/Queries/for_SerializedEmissionGate/when_emitting/and_the_gate_is_disposed.cs
+++ b/Source/DotNET/Arc.Core.Specs/Queries/for_SerializedEmissionGate/when_emitting/and_the_gate_is_disposed.cs
@@ -1,0 +1,31 @@
+// Copyright (c) Cratis. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Cratis.Arc.Queries.for_SerializedEmissionGate.when_emitting;
+
+/// <summary>
+/// The subscription tears the gate down before the underlying subject has stopped delivering. The emission that
+/// slips through must be dropped rather than throwing, because it runs from an async-void observer whose escaping
+/// exception would crash the process.
+/// </summary>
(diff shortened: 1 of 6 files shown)


woksin:
### Reviewer notes

Found by a bug-hunting pass over the streaming query pipeline (no associated GitHub issue).

**Root cause.** In `ObservableQueryDemultiplexer.SubscribeToSubjectOfType`, `OnEmission` is an `async void` Rx observer callback invoked from the MongoDB change-stream thread (no `SynchronizationContext`), so any escaping exception is unobserved and crashes the process. Two teardown races reached that:
- `await emissionGate.WaitAsync(token)` sat *outside* the `try`; the WebSocket teardown cancels the token before disposing subscriptions, so a parked emission threw `OperationCanceledException` out of the callback.
- The SSE teardown disposes subscriptions *before* cancelling the token, so an emission that passed the entry `IsCancellationRequested` guard hit a disposed `SemaphoreSlim` → `ObjectDisposedException` from `WaitAsync`/`Release`.

**Fix.** Extracted the gate into an internal `SerializedEmissionGate`: same one-at-a-time serialization, but it drops emissions (guarded `WaitAsync`, a post-acquire disposed re-check, and a guarded `Release`) instead of throwing once disposed or cancelled — in either teardown order. The demultiplexer now hands each emission to the gate rather than managing acquire/release inline.

**Tests.** 4 new `for_SerializedEmissionGate` specs (runs normally; dropped-without-throw when disposed; dropped-without-throw when cancelled; serializes overlapping emissions). Full `for_ObservableQueryDemultiplexer` suite green (46 total). Arc.Core Release build 0 warnings / 0 errors.

Not addressed here (separate follow-ups from the same pass): `IAsyncEnumerable`-backed observable subscriptions are never tracked (unsubscribe no-op / re-subscribe leak), and `QueryableExtensions.OrderBy` null-derefs on an unknown sort field (surfaces as 500 rather than 400).

woksin:

PR #2369, opened 2026-07-23
Fix serialization of float- and DateTime-backed concepts

woksin:
## Fixed

- A `float`-backed concept can now be persisted to MongoDB — it previously threw on every write
- A `DateTime`-backed concept now round-trips correctly instead of being read back as a wildly wrong date

### DIFF
--- a/Source/DotNET/MongoDB.Specs/for_ConceptSerializer/given/DateTimeConcept.cs
+++ b/Source/DotNET/MongoDB.Specs/for_ConceptSerializer/given/DateTimeConcept.cs
@@ -1,0 +1,8 @@
+// Copyright (c) Cratis. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Cratis.Concepts;
+
+namespace Cratis.Arc.MongoDB.for_ConceptSerializer.given;
+
+public record DateTimeConcept(DateTime Value) : ConceptAs<DateTime>(Value);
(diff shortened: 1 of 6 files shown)


woksin:
### Reviewer notes

Found by a bug-hunting pass over the MongoDB layer (no associated GitHub issue).

- **float** (`ConceptSerializer.cs`): `bsonWriter.WriteDouble((double)underlyingValue)` unbox-casts a boxed `float` directly to `double`, which throws `InvalidCastException` — unboxing requires the exact runtime type. Every sibling numeric branch unboxes to its exact stored type; only `float` crossed types. Fixed to `(double)(float)underlyingValue`. Deserialize was already correct (`Convert.ToSingle`), which is why the asymmetry pinpointed it.
- **DateTime** (`ConceptSerializer.cs`): write stored `dateTime.ToUniversalTime().Ticks / TimeSpan.TicksPerMillisecond` — ms since 0001-01-01 — while deserialize reads BSON DateTime via `DateTimeOffset.FromUnixTimeMilliseconds` (Unix epoch). A 2026 value came back around year 3994, and values near `DateTime.MaxValue` threw on read. Write now uses `ToUnixTimeMilliseconds()`, matching the read and the driver's own BSON DateTime convention. No migration concern: the old write never produced a value the read could recover, so no correctly-persisted data exists.

**Tests.** Added a `Roundtrip<T>` helper (serialize → deserialize) and round-trip specs for both; both fail against the old code. `for_ConceptSerializer` suite green (7 total). MongoDB Release build 0 warnings / 0 errors.

woksin:

PR #2370, opened 2026-07-23
Guard against a null event in the command response value handler

woksin:
## Fixed

- A command that returns an event collection containing a null element no longer fails with a server error during response detection — the collection is simply treated as not handled


woksin:
### Reviewer notes

Found by a bug-hunting pass over the command pipeline (no associated GitHub issue).

`EventsCommandResponseValueHandler.CanHandle` ran `objects.All(o => eventTypes.HasFor(o.GetType()))`, dereferencing each element with no null check. A handler returning e.g. `new object[] { someEvent, null }` threw `NullReferenceException` during the response-detection pass (inside the pipeline's `try`), surfacing as HTTP 500. The sibling `EventsForEventSourceIdCommandResponseValueHandler.CanHandle` already guards `item is null` — this brings the two into line: a null element makes the collection simply not handled here.

**Tests.** Added `with_a_null_element_in_the_collection` (returns false, no throw); it fails against the old code. `for_EventsCommandResponseValueHandler` suite green (51 total). Chronicle.Specs Release build 0 warnings / 0 errors.

woksin:

PR #2371, opened 2026-07-23
Resolve the aggregate root itself, not the Task the factory returns

woksin:
## Fixed

- Injecting an aggregate root into a command's `Handle` now resolves the aggregate root itself rather than failing to bind the argument to a `Task`

### DIFF
--- a/Source/DotNET/Chronicle.Specs/Aggregates/for_AggregateRootServiceCollectionExtensions/SampleAggregateRoot.cs
+++ b/Source/DotNET/Chronicle.Specs/Aggregates/for_AggregateRootServiceCollectionExtensions/SampleAggregateRoot.cs
@@ -1,0 +1,6 @@
+// Copyright (c) Cratis. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Cratis.Arc.Chronicle.Aggregates.for_AggregateRootServiceCollectionExtensions;
+
+public class SampleAggregateRoot : AggregateRoot;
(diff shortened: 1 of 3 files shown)


woksin:
### Reviewer notes

Found by a bug-hunting pass over the aggregate subsystem (no associated GitHub issue).

The scoped registration in `AggregateRootServiceCollectionExtensions.AddAggregateRoots` invoked `IAggregateRootFactory.Get<TAggregateRoot>` by reflection and returned the result directly. `Get` is `async` (it rehydrates the aggregate from its event stream), so `MethodInfo.Invoke` returns a boxed `Task<TAggregateRoot>`, and MS.DI stores that Task as the service instance. When a command declares `Handle(MyAggregate aggregate, …)`, the argument resolver injects the `Task<MyAggregate>` and `handleMethod.Invoke` throws `ArgumentException` — so every command injecting an aggregate root fails. The fix unwraps the Task, mirroring the existing `ReadModelServiceCollectionExtensions.ReleaseReadModel` pattern.

Latent today: there is no `AggregateRoot` subclass in the repo yet and no prior spec for this registration, so it was unexercised — it would break on first real use.

**Tests.** Added `for_AggregateRootServiceCollectionExtensions/when_resolving_a_registered_aggregate_root` (resolves the aggregate root, not a Task); fails against the old code. Chronicle.Specs Release build 0 warnings / 0 errors.

Note on CI: the repo's build workflow appears to cancel in-progress runs across PRs, so this PR's checks may need a maintainer re-run. Verified locally: Debug specs green, Release build of Chronicle + specs 0/0.

woksin:

PR #2372, opened 2026-07-23
Generate proxies for enums not backed by int

woksin:
## Fixed

- A command, query, or read model that references an enum not backed by `int` (for example `enum X : byte` or `[Flags] enum X : long`) no longer breaks TypeScript proxy generation

### DIFF
--- a/Source/DotNET/Tools/ProxyGenerator.Specs/for_TypeExtensions/when_converting_to_enum_descriptor/and_enum_is_byte_backed.cs
+++ b/Source/DotNET/Tools/ProxyGenerator.Specs/for_TypeExtensions/when_converting_to_enum_descriptor/and_enum_is_byte_backed.cs
@@ -1,0 +1,23 @@
+// Copyright (c) Cratis. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Cratis.Arc.ProxyGenerator.Templates;
+
+namespace Cratis.Arc.ProxyGenerator.for_TypeExtensions.when_converting_to_enum_descriptor;
+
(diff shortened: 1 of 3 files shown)


woksin:
### Reviewer notes

Found by a bug-hunting pass over the proxy generator (no associated GitHub issue), and reproduced under the generator's real `MetadataLoadContext` path.

`TypeExtensions.ToEnumDescriptor` did `Enum.GetValuesAsUnderlyingType(type).Cast<int>()`. `GetValuesAsUnderlyingType` returns the enum's actual underlying type, and `Cast<int>` unbox-casts each element to `int`, which throws `InvalidCastException` for any non-`int` underlying type (`byte`, `sbyte`, `short`, `ushort`, `uint`, `long`, `ulong`). That is an unhandled exception that aborts the whole generation run, so `dotnet build` fails for the consuming project. Every enum in the existing spec suite is `int`-backed, so the path was never exercised. `EnumMemberDescriptor.Value` is `object` and `AllFlagsExpression` uses `Convert.ToInt64`, so keeping the values boxed as their underlying type is sufficient — `int`-backed enums are unchanged (still a boxed `int`).

**Tests.** Added `for_TypeExtensions/when_converting_to_enum_descriptor` covering a `byte` enum and a `[Flags] long` enum (values preserved, all-flags expression correct); both fail against the old code. Ran with the worktree's JS workspace bootstrapped; `for_TypeExtensions` green (6 facts). ProxyGenerator + specs Release build 0 warnings / 0 errors.

Note on CI: the repo's build workflow appears to cancel in-progress runs across PRs, so this PR's checks may need a maintainer re-run. Verified locally as above.

woksin:

PR #2373, opened 2026-07-23
Map Uri to a TypeScript string, not Date

woksin:
## Fixed

- A `Uri` property now generates a TypeScript `string` instead of a `Date`, so its value is no longer silently lost to an invalid date on the client

### DIFF
--- a/Source/DotNET/Tools/ProxyGenerator.Specs/for_TypeExtensions/when_getting_target_type_for_uri.cs
+++ b/Source/DotNET/Tools/ProxyGenerator.Specs/for_TypeExtensions/when_getting_target_type_for_uri.cs
@@ -1,0 +1,14 @@
+// Copyright (c) Cratis. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Cratis.Arc.ProxyGenerator.for_TypeExtensions;
+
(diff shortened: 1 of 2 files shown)


woksin:
### Reviewer notes

Found by a bug-hunting pass over the proxy generator (no associated GitHub issue).

`TypeExtensions._primitiveTypeMap` mapped `Uri` to `_dateTargetType` (TS `Date`). Git history shows this was `new("string", "String")` and was changed to `_dateTargetType` inside commit `b37e0fa2` "Fixing constructor type for enums to be Number" — an accidental regression unrelated to that change. A `Uri` serializes to a JSON string, so the generated client's `new Date(url)` yields `Invalid Date`, which `JSON.stringify`s back to `null` — the value is silently lost, and the TS type is wrong for consumers. Mapped to the shared `_stringTargetType`, matching `char`/`string`.

**Tests.** Added `for_TypeExtensions/when_getting_target_type_for_uri` (maps to `string`, constructor `String`); fails against the old code. (The existing all-types scenario spec only asserted the Uri value was non-null, which is why the regression slipped through.) ProxyGenerator + specs Release build 0 warnings / 0 errors.

Note on CI: the repo's build workflow appears to cancel in-progress runs across PRs, so this PR's checks may need a maintainer re-run. Verified locally as above.

woksin:

PR #2374, opened 2026-07-23
Fix a batch of bugs found by a code audit

woksin:
## Fixed

- Header-based tenant resolution now applies for the whole request instead of only inside endpoint delegates, so tenant-scoped work that runs earlier no longer falls back to the default namespace and database; header lookup is now case-insensitive so a configured header still matches when HTTP/2 delivers it lowercased (#2351)
- An `IAsyncEnumerable`-backed observable query is now stopped on unsubscribe, instead of leaking and duplicating results on re-subscribe
- Observable queries served through the MongoDB watcher are now isolated per tenant database instead of all sharing the first tenant's
- The MongoDB watcher now resumes from the last change after a transient reconnect instead of silently dropping changes that happened during the disconnect
- A `float`-backed concept can now be persisted to MongoDB — it previously threw on every write
- A `DateTime`-backed concept now round-trips correctly instead of being read back as a wildly wrong date
- A command that returns an event collection containing a null element no longer fails with a server error during response detection
- Injecting an aggregate root into a command's `Handle` now resolves the aggregate root itself rather than a `Task`
- Executing a command with a base-type or interface result type now returns the result instead of throwing when the response is assignable to it
- An exception thrown by an aggregate root event handler now surfaces as itself, instead of being wrapped in a reflection exception that hid its type and stack
- An aggregate root commit result now reports only that aggregate's own events, with sequence numbers that correspond to them
- A command, query, or read model that references an enum not backed by `int` (for example `enum X : byte` or `[Flags] enum X : long`) no longer breaks TypeScript proxy generation
- A `Uri` property now generates a TypeScript `string` instead of a `Date`
- A dictionary whose key or value is a package type such as `Guid` or `TimeSpan` now generates compilable TypeScript, with the type's import included
- Command and query validation no longer relies on `Expression.Compile`, `Activator.CreateInstance` over a runtime-constructed generic, or private-member reflection, so it works under NativeAOT and trimming (#2204)


woksin:
### Reviewer notes

Seven bug fixes found by a code-audit pass, each an isolated commit with its own regression spec. Consolidated here (supersedes #2368, #2369, #2370, #2371, #2372, #2373, which are now closed). No associated GitHub issues.

1. **Observable emission crash** (`SerializedEmissionGate`) — `OnEmission` is an `async void` change-stream observer, so an escaping exception crashes the process. `WaitAsync` sat outside the `try` (OperationCanceledException on a cancel/teardown race) and the `SemaphoreSlim` gate could be disposed mid-emission (ObjectDisposedException). Extracted a gate that drops emissions instead of throwing once disposed or cancelled, in either teardown order.
2. **IAsyncEnumerable subscription leak** (`StreamingQuerySubscription`) — the async-enumerable branch started a background loop but returned `null`, so it was never tracked: unsubscribe couldn't stop it and re-subscribe started a second one. Now returns a disposable backed by a per-subscription linked CTS.
3. **`ConceptSerializer` float** — `(double)underlyingValue` unbox-cast a boxed `float` straight to `double` → `InvalidCastException` on every write. Unbox to `float` first.
4. **`ConceptSerializer` DateTime** — write stored ticks-since-0001 while read used `FromUnixTimeMilliseconds`; every value came back as a wildly wrong date. Store Unix-epoch ms.
5. **Events response value null guard** — `CanHandle` called `o.GetType()` on each element with no null check → NRE (HTTP 500). Matches the sibling handler's existing guard.
6. **Aggregate root DI** — the scoped factory returned the `Task<TAggregateRoot>` from the async `Get`, so injecting an aggregate into `Handle` failed to bind. Unwrap the Task, mirroring the read model resolver.
7. **Proxy generator, non-`int` enums** — `GetValuesAsUnderlyingType().Cast<int>()` threw for `byte`/`long`-backed enums, aborting generation. Keep values boxed as their underlying type.
8. **Proxy generator, `Uri`** — mapped to TS `Date` (an accidental regression from commit `b37e0fa2`), so a URL round-tripped to `null`. Mapped back to `string`.

**Verification (local).** Debug specs green for every touched area (Arc.Core.Specs Queries 561, Chronicle.Specs 53, MongoDB.Specs ConceptSerializer 7, ProxyGenerator for_TypeExtensions); Release builds of Arc.Core, Chronicle, MongoDB specs + ProxyGenerator source all 0 warnings / 0 errors. Each fix's spec fails against its pre-fix code.

Note on CI: the repo's build workflow appears to cancel in-progress runs across PRs, so checks here may need a maintainer re-run.

woksin:
### Reviewer notes — second batch (previously-deferred findings, now fixed)

These six came from the same audit and were deferred at first. Each was adversarially re-verified — a dedicated agent tried to refute it against the real code (and decompiled Chronicle types where needed) — before implementing, and each has a spec that fails on the pre-fix code.

- **Dictionary package-type import** (`TypeExtensions`): a dictionary with a `Guid`/`TimeSpan`/concept key or value emitted a `Record`/`ValueMap` referencing the type without importing it → uncompilable TS. The missing import is now emitted; existing dedup + reference-filtering keep it correct.
- **MongoDB watcher tenant isolation**: the process-wide singleton cached the first caller's database and served it to every tenant (cross-tenant read exposure with per-tenant DBs). The watch is now partitioned by resolved database name.
- **MongoDB watcher resume token**: reconnect re-opened the change stream with no resume position, dropping changes during the disconnect. It now carries and resumes from the last delivered token.
- **Aggregate handler exception unwrap**: `MethodInfo.Invoke` wrapped a synchronous throw from a void/non-async handler in `TargetInvocationException`; now unwrapped with `ExceptionDispatchInfo`, mirroring the command path.
- **`Execute<TResult>` polymorphism**: asking for a base type / interface the response is assignable to threw `InvalidCastException` (generic invariance). It now re-wraps when the response `is TResult`; the documented wrong-type throw is preserved (its existing spec still passes).
- **Aggregate commit Events/SequenceNumbers**: the result mixed unit-of-work-wide `Events` with aggregate-local `SequenceNumbers`; it now reports only the aggregate's own events and only derives sequence numbers when it owns the whole unit of work.

**Investigated and dropped (not a bug):** a suspected concurrency-violation-masking path in `AppendResultExtensions.ToCommandResult`. Decompiling the pinned Chronicle types showed `HasConcurrencyViolations` is derived from the very field the matched branches read, so the masking state is unreachable. No change made.

**Still open (documented, not fixed):** a narrow SSE subscribe/teardown subscription-leak race, and `QueryableExtensions.OrderBy` returning 500 rather than 400 on an unknown sort field (whether to skip or reject is a design call).

**Verification (local):** every touched area's Debug specs pass and Release builds are 0 warnings / 0 errors. CI on this repo cancels runs across PRs, so checks here may need a maintainer re-run.

woksin:
Folded in the former #2367: header-based tenant resolution (the ## Fixed bullet, #2351) plus the AOT quick-wins — removing runtime reflection from four validation sites (BaseValidator Expression.Compile, RuleBuilderInitialExtensions private reflection, DiscoverableValidators and ValidatorInvoker MakeGenericType). The AOT change is internal with no user-facing effect, so it is intentionally not in the release-note body. This bumps the PR to `minor` because it adds the public `HttpRequestContextMiddleware` type.

woksin:
Rebased onto main after #2375 merged. #2375 fixes the observable-emission async-void crash independently (inline exception guards), so this PR's `SerializedEmissionGate` and its specs were dropped in favor of main's fix, and the corresponding release-note bullet was removed. The unique `IAsyncEnumerable` subscription-tracking fix (`StreamingQuerySubscription`) is re-applied on top of main's demultiplexer, which #2375 does not address. Merged cleanly; the demultiplexer specs (main's disposed-emission specs + the async-enumerable spec) pass together.

woksin:
Ran a two-track adversarial review (correctness + spec-rigor) over the full diff. Verdict: correctness-sound, no blockers — all changes verified behavior-preserving or genuinely fixing the stated bug, and every behavioral fix has a spec that fails on the pre-fix code. Two low-severity edge cases surfaced by the review are now fixed in follow-up commits:

- **Aggregate event handling**: the `TargetInvocationException` unwrap enclosed the `await` of an async handler's task, so an async handler that itself faulted with a `TargetInvocationException` was mis-unwrapped to its inner exception. Now only `MethodInfo.Invoke` is guarded (`when_an_async_handler_faults_with_a_reflection_exception`).
- **MongoDB watcher**: `Lazy<DatabaseWatch>` (default `ExecutionAndPublication`) cached a factory exception permanently, so a transient `clientFactory.Create()` failure poisoned that tenant's watch for the process lifetime. The entry is now evicted on failure so a later `Observe` retries (`and_the_first_connection_attempt_fails`).

Also added a release-note bullet for the #2204 NativeAOT/trimming quick-wins in the validation pipeline, which the body had omitted. The validation refactor itself is behavior-preserving and guarded by existing specs (`for_BaseValidator/when_validating_concept_as_property` asserts the overridden property names). The two new watcher specs deliberately follow the existing plain-xUnit sibling (`and_the_change_stream_fails`) in that folder rather than the BDD base, for local consistency.

PR #2380, opened 2026-07-27
Restructure and fill out the Chronicle integration documentation

woksin:
## Added

- New "Use current state in a command" guide covering read model injection into a command's `CommandValidator<>`, `Provide()`, and `Handle()`, and what a nullable parameter means.
- New reference page for every read model resolution failure, with the cause and fix for each.
- Landing pages for the Chronicle overview and the tenancy section, neither previously reachable from navigation.

## Changed

- Chronicle read model documentation restructured into an overview, a command-injection reference, and a failure reference, with diagrams of how an instance is resolved from the command's key.
- Chronicle section reordered by workflow: set up, write, identity, read, validate, then advanced and reference topics.
- Aggregate pages renamed from "Aggregate Root" and "Aggregate Roots" to "Defining an aggregate root" and "Aggregate roots in commands".
- Tenancy, compliance, and reactor landing pages rewritten to explain the problem each one solves.
- Documented that a read model is injectable when backed by a projection, model-bound projection, or reducer, and that `[ReadModel]` alone is not sufficient.

## Fixed

- Aggregate roots were documented as transient services; they are command-scoped.
- Conflicting guidance on where a state-dependent validation rule belongs: race-prone invariants belong in `Handle()` or a Chronicle constraint, projected-state gates in a validator.
- Seven cross-links into the Chronicle documentation rendered as plain text instead of links.
- `/arc/backend/tenancy/` returned 404 because the section had no landing page.
- Chronicle validation page omitted `[Key]` and `ICanProvideEventSourceId` from the event source id conventions.
- Broken links on the proxy generation and frontend query pages.

## Removed

- Orphaned duplicate Chronicle commands page that was unreachable from navigation but still carried outdated content.


woksin:
Reviewer context (not release notes):

**Verification**
- `markdownlint-cli2` clean across all 258 Documentation markdown files.
- Every relative link and heading anchor in the Arc docs tree resolves. Four findings remain, all pre-existing on `main` and untouched here: two `proxy-generation` anchors, `frontend/core/queries.md -> ./commands.md`, and the DocFX `../../api/arc/` toc entry.
- Every framework API in the new examples was checked against source, not other doc pages: `ReadModelServiceCollectionExtensions`, the exception types (the documented client messages are their real `ValidationResult` text), and the `Chronicle.Specs/Commands/for_CommandScenario` fixtures, which have specs for validator, `Provide()`, and `Handle()` injection plus reducer-backed read models.

**Not verified**
- The full docs site build (`Documentation/web && npm run check`) was not run. The sync script resolves product repos as siblings of the `Documentation` repo, so it reads the main `Arc/` checkout rather than the worktree this branch was built in. Sidebar bucketing and rendered output are therefore unverified. No top-level toc section was added or renamed, so nav bucketing should be unaffected.
- Mermaid: the new `sequenceDiagram` parses clean. The new `flowchart` fails to parse in bare Node with `DOMPurify.addHook is not a function`, which is environmental rather than a syntax error: a minimal `flowchart TB / A --> B` passes while *any* quoted label fails, and `backend/chronicle/index.md` already ships a quoted-label flowchart today.

**Authorship** — these commits were written by Claude Code and merged without human review.

PR #2383, opened 2026-07-28
Resolve the default tenant to the unsuffixed read model database

woksin:
# Summary

A tenant named `Default` read from a database that does not exist, so every read model query silently returned no rows.

## Added

- `TenantId.Default` and `TenantId.IsDefault` for treating the default tenant — unset, or named — consistently across tenant-partitioned storage.

## Fixed

- A tenant named `Default` now resolves the unsuffixed MongoDB database, matching where Chronicle materializes the default namespace's read models. Previously it resolved `<database>+Default`, which does not exist, so both `IMongoCollection<TReadModel>` one-shot reads and `.Observe()` live queries returned no rows without erroring.

### DIFF
--- a/Source/DotNET/Arc.Core.Specs/Tenancy/for_TenantId/when_asking_whether_it_is_the_default.cs
+++ b/Source/DotNET/Arc.Core.Specs/Tenancy/for_TenantId/when_asking_whether_it_is_the_default.cs
@@ -1,0 +1,15 @@
+// Copyright (c) Cratis. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Cratis.Arc.Tenancy.for_TenantId;
+
+public class when_asking_whether_it_is_the_default : Specification
+{
+    [Fact] void should_consider_an_unset_tenant_the_default() => TenantId.NotSet.IsDefault.ShouldBeTrue();
+
+    [Fact] void should_consider_the_default_tenant_the_default() => TenantId.Default.IsDefault.ShouldBeTrue();
+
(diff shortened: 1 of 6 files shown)


woksin:
Reviewer context.

**Related:** Cratis/Chronicle#3522 centralizes and validates Chronicle's side of the same database naming, and documents the asymmetry this PR resolves against.

**Why the fix is here and not in Chronicle.** Both the injected `IMongoCollection<TReadModel>` (`ServiceCollectionExtensions`, the scoped `IMongoDatabase`) and `.Observe()` (`MongoDBWatcher`) resolve through `IMongoDatabaseNameResolver`, so one change covers both paths. Chronicle cannot register that resolver without inverting the Arc→Chronicle dependency direction, and changing Chronicle to suffix the default namespace instead would fix the `tenant = "Default"` case while breaking the `tenant = NotSet` case.

The full matrix, for reference — only the middle column was broken:

| | tenant `NotSet` | tenant `"Default"` | tenant `"Ada"` |
|---|---|---|---|
| `TenantNamespaceResolver` → namespace | `Default` | `Default` | `Ada` |
| Chronicle materializes read models into | `Ada` | `Ada` | `Ada+Ada` |
| `DefaultMongoDatabaseNameResolver` read (before) | `Ada` | `Ada+Default` | `Ada+Ada` |

**Behavior change to be aware of.** A pure-Arc MongoDB application (no Chronicle) that uses a tenant literally named `Default` will now read from the unsuffixed database rather than `<database>+Default`.

**Verification.** Full solution Release: 0 errors, 0 warnings (37 projects). Debug is also clean — the one CS0436 warning in `TestApps/Chronicle` reproduces on an unmodified `origin/main`, so it is pre-existing. Specs: Arc.Core 1552, MongoDB 139, Chronicle 568 — all pass.

PR #2384, opened 2026-07-28
Add Cratis.Arc.Screenplay package for generating Screenplay from Arc source

woksin:
# Summary

Generates a Cratis Screenplay `.play` document from the source of an Arc application.

## Added

- `Cratis.Arc.Screenplay` package that generates a Screenplay `.play` document describing an Arc application - concepts, policies, modules, features and slices, with the commands, events, queries, projections, reactors, constraints and screens in each
- Generation from source rather than from a running application, so a `.play` file can be produced from a checkout and regenerated in CI
- Generation from a Roslyn `Compilation`, so the package can be driven from a CLI, an MSBuild task or a source generator
- Diagnostics naming any construct that has no Screenplay counterpart, and where it was found
- Model-bound `[ChildrenFrom]` and `[Nested]` projections are recovered, matching what the fluent form already produced (#2385)
- Named authorization policies are recovered, with their requirements read from `AddPolicy` registrations where those can be read (#2386)
- Aggregate roots are followed from the command that reaches them, so a handler that mutates an aggregate states the events it produces
- Screens are declared for the component beside each slice

### DIFF
--- a/Arc.slnx
+++ b/Arc.slnx
@@ -30,6 +30,8 @@
         <Project Path="Source\DotNET\Swagger\Swagger.csproj" />
         <Project Path="Source\DotNET\OpenApi\OpenApi.csproj" />
         <Project Path="Source\DotNET\OpenApi.Specs\OpenApi.Specs.csproj" />
+        <Project Path="Source\DotNET\Screenplay\Screenplay.csproj" />
+        <Project Path="Source\DotNET\Screenplay.Specs\Screenplay.Specs.csproj" />
     </Folder>
     <Folder Name="/Source/DotNET/Tools/">
         <Project Path="Source\DotNET\Tools\ProxyGenerator\ProxyGenerator.csproj" />
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -19,6 +19,7 @@
     <PackageVersion Include="Cratis.Chronicle.Testing" Version="16.7.0" />
     <PackageVersion Include="Cratis.Fundamentals" Version="7.16.6" />
     <PackageVersion Include="Cratis.Metrics.Roslyn" Version="7.16.6" />
+    <PackageVersion Include="Cratis.Screenplay" Version="1.5.2" />
     <!-- MAUI -->
     <PackageVersion Include="Microsoft.Maui.Controls" Version="10.0.90" />
     <PackageVersion Include="Microsoft.Extensions.Logging.Debug" Version="10.0.10" />
--- a/Documentation/generating-a-screenplay.md
+++ b/Documentation/generating-a-screenplay.md
@@ -50,6 +50,26 @@ transported rather than what it is, so the query was left out (Library.Messaging
 
 Diagnostics come in three severities. **Information** means something is worth knowing but the document is complete. **Warning** means something was left out. **Error** means the document should not be trusted at all — the most common cause being source that did not compile.
 
+## Screens
+
+A [vertical slice](./vertical-slices.md) puts the React component that realizes a slice's screen in the same folder as its C#. Roslyn syntax trees carry real file paths, so the generator knows where each slice's source lives and declares a `screen` for every `.tsx` component sitting next to it, named after the file:
+
+```text
+slice StateChange Registration
+  command RegisterAuthor
+    name : AuthorName
+
+  screen AddAuthor
+    file Authors/Registration/AddAuthor.tsx
+```
+
+That is the whole of it — the `file` form and nothing else. The declarative form (`data … via query …`, `section`, `table`, `summary`, `action`, `navigate to`) is never inferred, because it would have to be read out of TypeScript rather than out of the compilation, and a guess at a screen's structure is worse than a pointer to the file that actually is it. Write those directives by hand if you want them, and expect a regeneration to leave them out.
+
+Two rules keep the result honest:
+
+- **Components only, no descending.** A file carrying a second extension — `AddAuthor.stories.tsx`, `AddAuthor.spec.tsx` — is a companion of a component, not a screen. A folder *inside* a slice folder is a slice of its own under the same convention, so its files belong to it.
+- **Anything uncertain is reported.** The relationship between a file and a slice comes entirely from where the file sits, so `SP0025` is reported whenever sitting there says less than usual: a slice whose source is spread over several folders, one folder holding the source of several slices, or two files claiming a single screen name.
+
 ## What is not expressed
 
 A `.play` is a description of an application, not a second copy of it. **It does not round-trip back to code**, and reading one will not tell you everything the code does. The losses run in both directions.
@@ -63,9 +83,8 @@ These are part of the language, but nothing in C# says them, so a generated docu
 | `capture` | Describes ingesting an external system. Nothing in an Arc application declares one. |
 | `persona` | Who uses the system is a product decision, not a code artifact. |
 | `seed` | Sample data is a modeling concern, not something the source states. |
-| `screen` | Screens are React components; their structure is not recoverable from C#. |
+| The declarative form of `screen` | What a screen shows and does is written in TypeScript and JSX. Only the `file` reference is generated — see [Screens](#screens). |
 | `@sensitive` | Only `[PII]` has a counterpart. Other sensitivity levels are not declared in Arc. |
-| Named authorization policies | `[Authorize(Policy = "…")]` names a policy whose rule lives in host configuration, not in the attribute. Roles are recovered; policy bodies are not. |
 
 ### Detail Screenplay cannot represent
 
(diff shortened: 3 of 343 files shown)


woksin:

**This is agent-authored code that no human has read yet.** 231 files, ~13.3k lines. The specs are green and the output is verified, but that is not a substitute for review — please treat the spec suite as a claim to be checked rather than evidence on its own. The areas I would look at first are `Analysis/` (is the symbol and body reading actually correct?) and whether the specs assert meaningful behavior rather than merely existing.

**Verification**

- `dotnet build Arc.slnx -c Release`: 0 errors. One warning, `CS0436` in `TestApps/Chronicle`, which is pre-existing and reproduces on `main` without this branch.
- `Screenplay` and `Screenplay.Specs` in Release across net8.0/net9.0/net10.0: 0 errors, 0 warnings.
- 417 specs passing. Each of the five commits builds on its own.
- End to end: a real Arc project (`TestApps/AspNetCore`) generates a `.play` that compiles under the **official** Screenplay compiler (`Cratis/Screenplay` `Source/DotNET/Tool`) with 0 errors and 0 warnings. That independent check is the one that matters; the round-trip helper inside the package is self-verification.

**Design note**

The core takes a Roslyn `Compilation` and nothing else — workspace loading stays in callers. That is what lets the same package serve a CLI, an MSBuild task, a source generator, and specs built from source strings with no mocks. `Model`, `Analysis` and `Emission` are separate, and `Model` has no Roslyn, Arc or Chronicle types in it.

**Known gaps, all reported by diagnostics rather than dropped silently**

- Model-bound `[ChildrenFrom]` / `[Nested]` are not read back (`SP0015`).
- `[Authorize(Policy = "X")]` reads as "requires authentication"; named policies beyond roles are not modelled.
- `Length(min, max)` splits into `Min` + `Max`, and a following `WithMessage` attaches only to `Max`.
- Not modelled at all: `capture`, `persona`, `seed`, `@sensitive`, `$strings.`, `import`, `authentication` providers. `screen` is now reachable — source analysis yields real file paths — but is not started.
- A project that does not restore currently yields a nearly empty document plus a warning and exit 0. That is arguably too quiet and worth hardening.

**Unverified risk**

Arc runs source generators. If any *discoverable artifact* is generated rather than hand-written, whether it appears depends on whether `MSBuildWorkspace.GetCompilationAsync()` includes generator output. Believed low-risk since `[Command]` and `[EventType]` declarations are hand-written, but I did not confirm it.

**Ordering**

The CLI side (`cratis screenplay generate`) is ready on a branch in `Cratis/cli` but is intentionally not opened yet: it must switch from a local project reference to a `PackageReference`, so it cannot merge until this package is published. Note also that the `cli` repo pinned `Cratis.Screenplay` 1.4.0 while this package compiles against 1.5.2 — misaligned versions fail assembly load at runtime.

woksin:
Two additions since the description was written, both now pushed.

**Documentation** — `Documentation/generating-a-screenplay.md`, wired into `toc.yml`. Written as an explanation page with a short how-to. It states in both directions what a generated document does *not* carry: Screenplay constructs that cannot be inferred from C# (`capture`, `persona`, `seed`, `screen`, `@sensitive`, named policies), and the Arc/Chronicle detail Screenplay has no way to represent (event generations, `[Tombstone]`, `[CompensationFor]`, reducer folds, aggregate roots, the event source id from `(TKey, TEvent)` handlers, read-model tags, query paging, custom routes). It says plainly that a `.play` does not round-trip back to code, and adds a "when this is the wrong fit" section for teams who maintain a `.play` as design-first intent. `verify-markdown.sh` passes.

**A fixed defect** — pointing the generator at a project that does not compile produced a nearly empty document, a warning and a *successful* exit. That now raises `SP0024` at Error severity, so a host exits non-zero. Whatever was recovered is still returned rather than discarded, and `SP0001` ("declares nothing") is suppressed in that case — reporting both would tell someone their application is empty when it is actually broken.

**Verification**

- 429 specs passing (was 417).
- `Arc.slnx` Release: 0 errors; the one warning is the pre-existing `CS0436` in `TestApps/Chronicle`, which reproduces on `main` without this branch.
- `Screenplay` / `Screenplay.Specs` Release across net8.0/net9.0/net10.0: 0 errors, 0 warnings.
- `Documentation/verify-markdown.sh`: all checks passed.
- End to end against the `Samples/Library` sample application: an 87-line `.play` that compiles under the **official** Screenplay compiler with 0 errors and 0 warnings, recovering the concepts, the module/feature/slice tree, projections and queries. Its diagnostics are the honest ones — `SP0013` for the event source id in a `(TKey, TEvent)` handler, `SP0015` for `Passive`, both of which Screenplay genuinely cannot express.

**Follow-ups filed rather than folded in**, to keep this reviewable: #2385 (model-bound `[ChildrenFrom]` / `[Nested]`) and #2386 (named authorization policies). The `Length(min, max)` message-attachment quirk is noted in the docs page.

woksin:
Three further gaps closed in `82d280d`, so the package no longer defers them.

**Model-bound `[ChildrenFrom]` / `[Nested]` (#2385).** Closing this uncovered a real silent drop, not just a missing feature: attributes on a positional record bind to the constructor *parameter*, and only the property was being read — so on the canonical Chronicle style these were discarded with no diagnostic at all. Event property references are now camel-cased too, so the model-bound and fluent styles describe the same projection identically rather than emitting `key SliceId` against `key sliceId`.

**Named authorization policies (#2386).** Names survive instead of collapsing to plain authentication, and requirements are recovered from `AddPolicy` registrations where they can be read; where they cannot, that is reported rather than guessed. This also fixed a latent bug — a lowercase role emitted `authorize librarian`, which the Screenplay parser rejects, so the generator could produce a document that would not compile.

**Aggregate roots.** Followed from the command that reaches them, so a handler that mutates an aggregate states the events it produces instead of appearing to produce none. `SP0018` is narrowed rather than removed: it now fires only for an aggregate root no command reaches, which is the only case where the events genuinely go unstated.

**Screens.** Declared in their `file` form, pointing at the component beside each slice. The declarative form (`data … via query`, `table`, `column`) needs a TypeScript syntax tree and is deliberately not inferred. Filesystem access sits behind an injected abstraction so the specs stay hermetic.

**Verification**

- 603 specs passing (was 429).
- `Arc.slnx` Release: 39 projects, 0 errors, 0 warnings.
- `Screenplay` / `Screenplay.Specs` Release across net8.0/net9.0/net10.0: 0 errors, 0 warnings.
- `Documentation/verify-markdown.sh`: all checks passed.
- End to end against the `Samples/Library` sample application: a 99-line `.play` (was 87) that compiles under the **official** Screenplay compiler with 0 errors and 0 warnings.

**One thing worth flagging for review.** `82d280d` is a single commit covering three concerns, which is against the usual one-logical-unit rule. They were built concurrently and share files — `ScreenplayDiagnosticCodes`, `ApplicationSyntaxBuilder`, `SliceReader`, the documentation page. I tried to split them and verified that the split does not build (`ApplicationSyntaxBuilder` carries an entangled `AuthorizeSyntaxBuilder` signature change), so I chose one commit that compiles over three that do not. The seven earlier commits were each verified to build on their own.

**Still not expressed, because there is nothing to infer them from:** `capture`/CDL has no Chronicle counterpart; `persona`, `seed`, `@sensitive` and `$strings.` have no source in an Arc backend. These are not deferred work.

The CLI (`cratis screenplay generate`) cannot be part of this PR — it lives in `Cratis/cli`. It is committed on a branch there and still cannot merge until this package is published, since it must switch to a `PackageReference`.

woksin:
Final round. An adversarial review with mutation testing, plus the last two recoverable features.

**Nine defects found and fixed.** Two would have reached a user:

- A string containing a newline split its construct across lines, and the document no longer compiled. Same shape as the `authorize librarian` bug.
- Generated `file` paths were anchored on the common prefix of *every* syntax tree. A real MSBuild compilation includes NuGet content files from `~/.nuget/packages`, which collapses that prefix to the filesystem root — so every path came out as the generating machine's layout, defeating the point of a document meant to be committed and diffed. **Only running the real generator against a real project found this**; no hermetic spec could.

The rest were silences: `[PII]` on a positional record parameter (the document said "not sensitive" while Chronicle encrypts the field), non-public reactor handlers, and inherited event properties were all dropped with no diagnostic; query discovery did not follow the rules Arc itself applies, so static helpers were stated as endpoints while real queries went missing; and partial types took their order from the compilation, so the same source could describe itself two ways.

Three of those silent drops share one cause — attributes and members on positional records — which is now handled in one place rather than three.

**On the spec suite, honestly.** Mutation testing ran 145 mutants across 70 implementation files: **40 survived (37%)**, and six of the nine behavioral fixes above left all 603 specs green. The survivors cluster in ordering and first-wins choices rather than core logic, so this is under-detection of future regressions rather than present incorrectness — but "690 green specs" should be read with that in mind, not as a guarantee. A document-level determinism spec was added, which would have caught the partial-types defect.

**Screens now carry `data <T> via query <Q>`.** Deliberately nothing else — no `table`, `column`, `section` or `action`. A `data` binding takes nothing from the component: an import only narrows the queries the slice already declares in C#, and everything stated comes from the model. An `action` would assert that the screen offers a command to a user, which an import cannot establish. A plausible invention is worse than an omission in a document whose only value is describing the real application.

**Conditions inside aggregate behaviors** resolve back to command input. A guard on aggregate *state* cannot be expressed as a comparison against command input, so it is reported (`SP0027`) rather than quietly restated as unconditional.

**Verification**

- 690 specs passing, Debug and Release.
- `Arc.slnx` Release: 39 projects, 0 errors, 0 warnings.
- `Documentation/verify-markdown.sh`: passed.
- End to end: real Arc source → CLI → `.play` compiling under the **official** Screenplay compiler with 0 errors, and every `file` path relative — the path defect verified fixed against the real generator, not just in specs.

**Worth adopting:** an end-to-end gate in CI that generates from a real sample and compiles the result with the Screenplay tool. Two of the nine defects were invisible to the hermetic specs and only appeared when the real generator met real source.

Remaining constructs — `capture`/CDL, `persona`, `seed`, `@sensitive`, `$strings.`, `import`, and the declarative screen widgets — are not deferred work. There is nothing in an Arc backend to infer them from.

PR #2391, opened 2026-07-28
Compose Chronicle update release notes and supersede older updates

woksin:
## Changed

- Chronicle dependency updates now open with release notes composed from the Chronicle releases they pick up, folded into the matching `Added`/`Changed`/`Fixed` sections, so the resulting Arc release describes what actually changed
- The verification gate for a Chronicle update is reported as a pull request comment instead of in the body, keeping CI bookkeeping out of published Arc release notes
- Only the newest Chronicle update stays open — older ones and any the `main` branch has already passed are closed with a comment explaining why

## Fixed

- Auto-merge no longer merges a Chronicle update that a newer one has superseded

### DIFF
--- a/.github/scripts/compose-chronicle-release-notes.py
+++ b/.github/scripts/compose-chronicle-release-notes.py
@@ -1,0 +1,196 @@
+#!/usr/bin/env python3
+"""Compose the pull request body for a Chronicle dependency update.
+
+The body doubles as Arc's release notes - `cratis/release-action` publishes the merged
+pull request's body verbatim as the GitHub release - so it has to follow
+`.github/pull_request_template.md` exactly: an optional short summary plus the
+`Added`/`Changed`/`Fixed`/`Removed`/`Security`/`Deprecated` sections that apply, and
+nothing else. Anything about CI belongs in a pull request comment, never in the body.
+
+Chronicle's own release notes follow the same template, so every Chronicle release picked
+up by the update contributes its bullets to the matching section here, attributed to the
+release it came from. Issue references are rewritten to point at Chronicle's issue tracker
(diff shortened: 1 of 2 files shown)


woksin:
### Why this was needed

`cratis/release-action` publishes a merged pull request's body verbatim as the GitHub release, so the verification gate table and the auto-merge verdict were landing in Arc's published release notes on every Chronicle update.

Separately, Arc had accumulated open Chronicle updates for 16.8.0 and 16.9.0 while `main` had already moved to 16.9.1 — all of them redundant, none of them closed.

### Verification

The workflow has not run in Actions yet; everything below is local.

- Release-notes composition run against the real 16.7.0 → 16.9.1 gap, plus unknown-previous version, API-failure fallback, wrapped and nested bullets, issue-reference rewriting, and prerelease/out-of-range filtering.
- Supersede logic executed against a mocked `gh` across five scenarios, including semantic ordering (`16.10.0` beating `16.9.1`, where a lexical sort picks wrong) and an out-of-order run closing its own pull request.
- Workflow YAML parses and all 12 `run` blocks pass `bash -n` after YAML dedenting; the gate comment renders correctly in all four outcome branches.
- `gh pr close --comment` and `create-pull-request@v7`'s `body-path` confirmed against their sources.

### Not addressed here

- Pull requests #2382 and #2387 are still open and redundant. The new step only runs on the next Chronicle publish, so they need closing by hand or via a `workflow_dispatch` run with version `16.9.1`.
- Chronicle's `notify-arc` job still dispatches only `{"version": …}`, so every Arc update is labeled `patch` regardless of Chronicle's own semver bump. A Chronicle major would currently auto-merge as an Arc patch.

Issue #2401, opened 2026-07-28
Screenplay generation hardening: small analysis and emission defects from review


woksin:
## Summary

A code review of `Cratis.Arc.Screenplay` (as merged in PR #2384) surfaced a set of smaller, high-confidence defects and gaps. Filing them together since each is a few lines; happy to split any of them out.

## Findings

1. **`MarkAsPii` does not unwrap collection elements the way `Resolve` does** (`Analysis/Types/TypeRegistry.cs`). `Resolve` re-unwraps a collection's element type; `MarkAsPii` does not, so a `[PII] IEnumerable<Name?>`-shaped member registers PII under `Nullable` rather than under the concept. Share the unwrap path between the two.

2. **Commented-out imports still produce screen-binding candidates** (`Analysis/Screens/ScreenImports.cs`). `StatementRegex` anchors on `^[ \t]*import`, so `// import { AllAuthors } from './Listing';` matches, and when it names a query the slice declares it becomes a false `data` binding. Strip `//` and `/* */` regions before matching.

3. **Query parameter filtering excludes only interface-typed parameters** (`Analysis/Queries/QueryReader.cs`, `IsInput`). Struct infrastructure parameters (`CancellationToken`, paging/sorting value types) would be emitted as caller input with an undeclared type name. Filter the known infrastructure types explicitly.

4. **`SP0029` is skipped in `ScreenplayDiagnosticCodes`** with nothing documenting why (SP0028 jumps to SP0030). If it was retired pre-merge, reserve it with a comment so it is not reused with a different meaning later.

5. **Unicode normalization order in `ScreenplayNaming.Sanitize`** (`Emission/Naming/ScreenplayNaming.cs`): characters are stripped *before* normalizing to Form C, so a decomposed `é` (e + combining accent) loses its accent while a precomposed `é` survives — canonically equal source produces different documents. Normalize first, then strip.

6. **No spec covers optional/nullable emission end-to-end.** `TypeRegistry.Unwrap` distinguishes `Nullable<T>` and annotated reference types, but no spec asserts that a nullable property comes out as `Type?` in the printed document. (Observed to work in practice — Ada emits `groupKey InvoiceGroupKey?` — but it is unspecified.)

[...]

Issue #2393, opened 2026-07-28
Screenplay generation: SP0024 reports 'nothing recovered reliably' and fails hosts when only source-generated symbols are missing


woksin:
## Summary

When the analyzed compilation depends on source-generator output, `ScreenplayGenerator` raises `SP0024` at **Error** severity with the message "Nothing recovered from it describes the application reliably" — even when recovery is in fact near-complete. Hosts that follow the documented contract (exit non-zero / discard output on `IsSuccess == false`) then throw away a document that is essentially correct.

## Evidence (real application)

Generating from a real Arc application (Ada, ~1,800 syntax trees) via `MSBuildWorkspace.GetCompilationAsync()`:

- The compilation had **607 errors**, first being `The name 'AccountsMessages' does not exist in the current context` — all of them missing *source-generated* types (message-constant classes, logging generators). `MSBuildWorkspace.GetCompilationAsync()` does not run source generators, so any app that relies on them fails to compile in this loading mode. This was flagged in PR #2384 as an unverified risk; it is now confirmed.
- Despite `SP0024 [Error]`, the generated document was **5,362 lines** and matched a previously generated known-good document for the same application almost byte-for-byte (the only diff was content that had legitimately changed in the source since). Structure, concepts, events, commands, produces mappings, projections (joins, children, composite keys), reactors, constraints, policies, and screens were all recovered correctly.
- The document compiles with the official Screenplay compiler: **0 errors, 1 warning**.

So the diagnostic's claim ("nothing … describes the application reliably") was wrong by roughly 5,300 lines, and its Error severity caused the host CLI (`cratis screenplay generate`, branch `feature/arc-screenplay` in Cratis/cli) to write **no output at all** for a real application.

Note that this will hit *most* conforming Arc applications, not edge cases: the Cratis house style itself relies on source generation (`[LoggerMessage]` partial classes, message-constant generators), so "compilation has missing generated symbols when loaded without running generators" is the common case, not the exception.

## Suggested fix

Two independent parts:

1. **Calibrate SP0024.** The generator already knows how much it recovered. Suggested shape:
   - Error only when the compilation errors actually prevented recovery (e.g. zero or near-zero artifacts recovered, or errors located inside discovered artifact declarations).
   - Otherwise a Warning that states both facts honestly: N compilation errors, M artifacts recovered, first error message, and a hint that missing types named like `*Messages`/generated classes usually mean source generators were not run by the host.
[...]

Issue #2386, opened 2026-07-28
Screenplay generator collapses named authorization policies to plain authentication


woksin:
## Problem / motivation

`Cratis.Arc.Screenplay` reads `[Authorize]` and `[Roles("X")]` into Screenplay `policy` declarations, but `[Authorize(Policy = "SomePolicy")]` is currently flattened to "requires authentication" with no role, so the named policy is lost from the generated `.play` document.

Screenplay's `policy` construct can express more than roles — it supports `claim <name> matches <value>` and boolean combinations — so a named ASP.NET policy has somewhere meaningful to go.

## Proposed solution

Model named authorization policies rather than collapsing them. The difficulty is that an ASP.NET policy's *definition* lives in `Program.cs` service registration rather than on the attribute, so the requirements behind the name may not be recoverable. At minimum the policy name should survive into the document as a referenced `policy` declaration; recovering its requirements from `AddAuthorizationBuilder`/`AddPolicy` registrations would be a further step worth evaluating separately.

## Acceptance criteria

- [ ] `[Authorize(Policy = "X")]` on a command or query yields an `authorize` referencing a policy named after `X`, not a generic authenticated policy.
- [ ] A `policy` declaration is emitted for it so the document compiles without a dangling reference.
- [ ] Where the requirements behind the policy cannot be recovered, that is reported as a diagnostic rather than silently guessed.
- [ ] Hermetic specs covering the named-policy form alongside the existing role forms.

## Additional context

Introduced with the package in #2384.

woksin:
## Resolved on `main` — and beyond the acceptance criteria

`Source/DotNET/Screenplay/Analysis/AuthorizationReader.cs` now reads the `Policy` named argument:

```csharp
public const string PolicyProperty = "Policy";
...
var policies = attributes
    .Select(_ => _.GetNamedArgument(PolicyProperty) as string)
    .OfType<string>()
    ...
return new(true, roles) { Policies = policies };
```

and `Analysis/Policies/PolicyCatalog.cs` declares them so the document has no dangling reference.

The "further step worth evaluating separately" in the issue — recovering the requirements behind the name from `AddAuthorizationBuilder` / `AddPolicy` registrations — also landed: `Analysis/Policies/PolicyRegistrations.cs`, `PolicyRegistration.cs`, `PolicyRequirementReader.cs` and `PolicyValues.cs`.

@einari @woksin — safe to close.

woksin:
[...]

Issue #2385, opened 2026-07-28
Screenplay generator does not read model-bound [ChildrenFrom] and [Nested] projections


woksin:
## Problem / motivation

`Cratis.Arc.Screenplay` reads model-bound projections from `[ReadModel]` types, but does not read back the child and nested collection attributes. A read model using `[ChildrenFrom]` or `[Nested]` currently emits an `SP0015` diagnostic and the child structure is absent from the generated `.play` document, even though Screenplay's projection language can express it (`children` and `nested` blocks).

The fluent `IProjectionFor<T>` equivalents (`.Children(...)`, `.Nested(...)`) **are** read, so this is an inconsistency between the two projection styles rather than a limit of the target language.

## Proposed solution

Read `[ChildrenFrom]` and `[Nested]` in the model-bound projection reader and populate the existing `ProjectionChildScopeModel`, which the emission half already consumes for the fluent path. No new model shapes should be needed.

## Acceptance criteria

- [ ] A `[ReadModel]` using `[ChildrenFrom]` produces a `children` block equivalent to the fluent form.
- [ ] The same for `[Nested]`.
- [ ] `SP0015` is no longer raised for these two attributes.
- [ ] Hermetic specs (C# source strings in, model asserted) covering both, plus a round trip proving the document compiles.

## Additional context

Introduced with the package in #2384. The gap is documented in `Documentation/generating-a-screenplay.md`.

woksin:
This appears to already be implemented on `main`. Commit `82d280d` ("Recover children, named policies and screens from source") added `ModelBoundChildren.In`/`NestedIn`, which read `Cratis.Chronicle.Projections.ModelBound.ChildrenFromAttribute` and `NestedAttribute` into `ProjectionChildScopeModel` via `ModelBoundScopeReader` — reusing the same emission path as the fluent `.Children()`/`.Nested()`, and no longer raising `SP0015` for those attributes. Hermetic specs exist (`a_model_bound_projection_with_children.cs`, `a_model_bound_projection_with_a_nested_object.cs`) plus a round-trip generator spec (`from_source_declaring_children_and_nested_objects.cs`).

@einari — looks safe to close as already resolved unless there's an aspect not covered by `82d280d`.

---
_Generated by [Claude Code](https://claude.ai/code)_

woksin:
## Resolved on `main`

[...]

PR #2402, opened 2026-07-28
Compile every generated Screenplay document before returning it

woksin:
## Added

- Generated documents are compiled before they are returned; one the Screenplay language rejects is reported as an error naming the first rejected line, so a host exits non-zero instead of writing a document that does not compile


woksin:

**Why this exists.** `RoundTrip` already printed a document, compiled it and reprinted it, and its own comment called it "the correctness gate for everything the generator produces". It lived in `Screenplay.Specs`, so it only ever ran against hand-written fixtures — the generator running against a real application verified nothing. That is exactly how a command property named `Description` shipped a document that did not compile: no fixture happened to use that name, and 690 green specs said nothing about it.

**Three decisions worth checking**

- **The verifier is held concretely, not injected.** `ScreenplayGenerator` has a private `ScreenplayVerifier` field rather than a constructor parameter. The check was decided always-on, and a host that could substitute it could switch it off. This also leaves the public surface of `ScreenplayGenerator` unchanged, so the published package is not broken.
- **Runtime does compile only.** Reprinting and comparing byte-for-byte stays a spec-level gate. Compiling is the cheap, decisive test of "is this valid"; byte-identical reprint is a stronger claim about information loss and belongs where the fixtures are known.
- **`IScreenplayVerifier` has no consumer that injects it.** Kept for consistency with every other capability in the package (`IScreenplayGenerator`, `IApplicationModelAnalyzer`, `IScreenplayEmitter`), and because adding a public interface later is additive while removing one is breaking. If you would rather not ship an abstraction nothing injects, deleting the file and dropping `: IScreenplayVerifier` is the whole change.

`SP0024` suppresses the new `SP0034`: a model recovered from source the compiler never accepted describes an application that does not exist, so a poor document made from it is a consequence already reported rather than a second defect. That suppression is covered by a spec which was confirmed to fail when the guard is removed.

**What this does not catch.** Compiling your own output catches "invalid", never "valid but untrue". The enum defect fixed earlier in #2384 emitted `role = 6`, which compiles perfectly and is simply wrong. An end-to-end CI gate that generates from a real project remains worth adding for that class.

**Verification**

- `Screenplay.csproj` Release with `bin`/`obj` deleted first: 0 errors, 0 warnings.
- 804 specs passing (783 before).
- Against a real third-party application (~5300 lines, 121 commands, 172 events): the new check reported **nothing** — 0 `SP0034`, 0 error-severity diagnostics, and the document still compiles under the official Screenplay tool with 0 errors. Generation remains deterministic, byte-identical across two runs.
- `TestApps/AspNetCore` end to end through the CLI: 52 lines, compiles 0 errors / 0 warnings.

No issue reference in the description because none of the open Screenplay issues (#2385, #2386, #2393–#2401) covers this.

Issue #2398, opened 2026-07-28
Screenplay generation: separator-carrying constraint names are sanitized into unreadable identifiers


woksin:
## Summary

Constraint names that carry separators in their runtime name (kebab-case is idiomatic for Chronicle constraint names) are sanitized by *deleting* the separator characters, producing unreadable all-lowercase run-together identifiers in the generated document.

## Evidence (real application)

Ada declares:

```csharp
public class UniqueTimesheetStartConstraint : IConstraint
{
    public const string Name = "unique-timesheet-start";

    public void Define(IConstraintBuilder builder) => builder
        .Unique<TimesheetStarted>(name: Name);
}
```

The generated document emits:

```
constraint uniquetimesheetstart
  unique event TimesheetStarted
```

All 31 constraints in the document look like this (`uniqueinvitationemail`, `uniqueengagementonboardingcompletion`, `uniquetimesheetoverduereminder`, …). The Screenplay identifier rules (`[A-Za-z_]\w*`) force *some* transformation, but deleting separators is the least readable choice, and it discards word boundaries that were present in the source.

## Suggested fix

When sanitizing a name for an identifier position, treat separator characters (`-`, `_`, spaces, dots) as word boundaries and PascalCase the segments: `unique-timesheet-start` → `UniqueTimesheetStart`. Where a declaring type exists (as with `IConstraint` implementations), the type name minus a `Constraint` suffix is an even better candidate (`UniqueTimesheetStartConstraint` → `UniqueTimesheetStart`).

Round-trip note: the constraint's *runtime* name is the string, and a future .play→code path would regenerate a different string from the PascalCase identifier. If that matters, the mapping convention (PascalCase ↔ kebab-case) should be stated in the docs.

## Environment

- `Cratis.Arc.Screenplay` 20.65.0 (`Emission/Naming/ScreenplayNaming.cs` sanitization)

woksin:
## Resolved on `main`

Landed in #2418:

> Names carrying separators read as words instead of running together, so a constraint named `unique-timesheet-start` is written as `UniqueTimesheetStart` (#2398)

`Source/DotNET/Screenplay/Emission/Naming/ScreenplayNaming.cs` splits on separators via `WordsIn` and joins in Pascal case, and leaves a name carrying no separator exactly as it is. The round-trip note from the issue is captured in the method's `<remarks>`.

@einari @woksin — safe to close.

woksin:
Closing — released in **v20.65.2**.

[...]

Issue #2395, opened 2026-07-28
Screenplay generation: obj/ generator-output folders pollute slice source-folder attribution (false SP0025)


woksin:
## Summary

Slice source-folder attribution includes syntax trees that live under `obj/` (compiler-generated files emitted to disk, e.g. `obj/Debug/net10.0/Microsoft.Gen.Logging/...`). This produces false `SP0025` diagnostics and, worse, widens the folder set that screen discovery scans ("every user interface file in any of them is taken to be a screen of it").

## Evidence (real application)

Generating from the Ada application:

```
SP0025 [Warning] The source of the slice is spread over 2 folders (/…/Ada/Source/Core/Onboarding/Registry, /…/Ada/Source/Core/obj/Debug/net10.0/Microsoft.Gen.Logging/Microsoft.Gen.Logging.LoggingGenerator), so every user interface file in any of them is taken to be a screen of it
SP0025 [Warning] The source of the slice is spread over 2 folders (/…/Ada/Source/Core/Requests/Candidates/AnalyzeCv, /…/Ada/Source/Core/obj/Debug/net10.0/Microsoft.Gen.Logging/Microsoft.Gen.Logging.LoggingGenerator), so every user interface file in any of them is taken to be a screen of it
```

The second folder in each case is not "source of the slice" in any meaningful sense — it is generator output that happens to declare partial-class members in the slice's namespace. The slice is *not* actually spread over two folders, so the diagnostic is misleading, and the screen scan now includes an `obj/` tree (harmless today because no `.tsx` lives there, but the attribution is wrong by construction).

## Suggested fix

When computing the folders a slice's source lives in (and the majority-vote project root in `SourcePaths`), exclude syntax trees whose paths contain an `obj/` (or `bin/`) segment relative to the project root — or more precisely, trees marked as generated (`.g.cs`, generator-emitted hint paths) — before attribution. Generated members contribute symbols to the slice but say nothing about where the slice's screens live.

## Environment

- `Cratis.Arc.Screenplay` 20.65.0, loaded via `MSBuildWorkspace` from a project with `EmitCompilerGeneratedFiles`-style generator output present under `obj/`

woksin:
## Resolved on `main`

Landed in #2418:

> Source written by the build no longer counts as part of a slice, removing incorrect reports of a slice being spread over several folders (#2395)

The exclusion is `Source/DotNET/Screenplay/Analysis/GeneratedSource.cs`:

```csharp
[...]

Issue #2394, opened 2026-07-28
Screenplay generation: emit 'import' for events declared outside the compilation instead of leaving dangling references (SP0021)


woksin:
## Summary

When an artifact references an event that is declared **outside the analyzed compilation**, the generator reports `SP0021` ("'X' is referred to but declared outside the compilation, so the document refers to an event it never introduces") — but then emits a document that indeed refers to an event it never introduces. The Screenplay language already has the construct for exactly this situation: `import`. The `ScreenplayCompiler` treats the short name of an `import` as a *known event*, which is precisely the semantic needed here.

## Evidence (real application)

Generating from the Ada application (whose reactors observe events published by a sibling bounded context):

```
SP0021 [Warning] 'InvitationToJoinAdaAccepted' is referred to but declared outside the compilation, so the document refers to an event it never introduces
```

The emitted document contains:

```
reactor AcceptedInvitationProvisioner
  on InvitationToJoinAdaAccepted
    file Admin/Invitations/Provision/Provision.cs
```

and compiling that document with the official Screenplay compiler yields:

```
Ada.play(785,11): warning: Unknown event 'InvitationToJoinAdaAccepted' - declare it with 'event InvitationToJoinAdaAccepted'
```

So the one warning the generated document produces under the official compiler is one the generator had all the information to avoid.

## Suggested fix

When an event symbol is resolved but lives outside `compilation.Assembly`, emit a top-level `import <Namespace.EventName>` (the compiler registers the trailing segment as a known event) instead of — or in addition to — SP0021. SP0021 can stay for the truly unresolvable case (symbol not found at all).

This also improves the story for multi-context applications: cross-context translation slices (an Arc app reacting to another context's public events) are common, and `import` is the DSL's way of stating that dependency explicitly.

## Environment

- `Cratis.Arc.Screenplay` 20.65.0
- `Cratis.Screenplay` compiler 1.5.2 (import handling: `ApplicationSyntax` known-events; validator warns on unknown reactor triggers)

woksin:
## Resolved on `main`

Landed in #2418:

> Events declared outside the application being read are imported by name, so a generated document no longer refers to an event it never introduces (#2394)

SP0021 is retained for the genuinely unresolvable case, as the issue proposed.

@einari @woksin — safe to close.

woksin:
Closing — released in **v20.65.2**.

Delivered by #2418. Verification against `main` is in the comment above.

PR #2406, opened 2026-07-28
Import external events, ignore build output in slices, and read separators in names

woksin:
## Added

## Fixed

### DIFF
--- a/.github/workflows/dotnet-build.yml
+++ b/.github/workflows/dotnet-build.yml
@@ -176,6 +176,57 @@ jobs:
       - name: Test (${{ matrix.project }})
         run: dotnet test ${{ matrix.project }} --no-build --configuration Release --framework net10.0
 
+  # Every Screenplay specification builds its compilation from source strings, which is what makes them hermetic -
+  # and what puts a whole class of defect out of their reach by construction. A compilation built that way has no
+  # intermediate folder, no source generator output on disk and no referenced package declaring an event, so a
+  # generator that mis-attributes one of those produces a wrong document that every specification still passes. Two
+  # shipped that way. This reads a real project through MSBuild and holds the document it generates to the compiler
+  # the language ships - it has to come back with nothing to say, warnings included.
+  screenplay-end-to-end:
+    needs: [dotnet-build]
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event_name == 'pull_request_target' && github.event.pull_request.head.sha || github.sha }}
+
+      - name: Setup .Net
+        uses: actions/setup-dotnet@v4
+        with:
+          dotnet-version: |
+            ${{ env.DOTNET8_VERSION }}
+            ${{ env.DOTNET9_VERSION }}
+            ${{ env.DOTNET10_VERSION }}
+
+      - name: Cache NuGet packages
+        uses: actions/cache@v6
+        with:
+          path: ~/.nuget/packages
+          key: nuget-${{ runner.os }}-${{ hashFiles('**/*.csproj', '**/Directory.*.props', '**/Directory.*.targets') }}
+          restore-keys: |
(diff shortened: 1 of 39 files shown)


woksin:

**This is agent-authored code that no human has read yet.** Green CI is not sign-off. Please treat the spec suite as a claim to be checked. **Do not merge without explicit review.**

**The most important thing here is not in the release notes.** `Source/DotNET/Screenplay.EndToEnd/` is a new console harness, wired into `dotnet-build.yml`, that loads `TestApps/AspNetCore` through a real MSBuild workspace, generates a document, and compiles it with the `Cratis.Screenplay` compiler. It fails on **any** compiler diagnostic about the generated document, warnings included — a warning means the generator wrote a document referring to something it never introduces, which is a defect here rather than in the application. This is `#2401` item 8, and it is the gate that catches the class of defect no specification built from source strings can reach: two of the nine defects found before this package first shipped were invisible to hermetic specs *by construction*, and `#2394`'s dangling-event warning is a third.

**A deliberate deviation on `#2398`.** The issue also asks to prefer the declaring type name minus a `Constraint` suffix. That was **not** adopted. A single `IConstraint` may declare several rules in one `Define`, and naming them all after the type collapses them into duplicate `constraint` names within a slice; it also discards the runtime name, which is the thing the document is describing. On the real application it buys nothing — every constraint's runtime string already PascalCases to exactly the type name minus `Constraint`. Easy to add narrowly if a reviewer disagrees.

**A limitation of the gate, stated plainly.** The committed harness does not carry the resx `PrepareResources` workaround that the CLI has, so it would still trip over an application whose resource classes are generated during the build. `TestApps/AspNetCore` has none, so CI is unaffected — but if the gate should ever point at such an application, that fix needs an equivalent here. Verifying against the real production application required setting `CustomAfterMicrosoftCommonTargets` externally; no file in either repository was changed for it.

**Verification**

- `Screenplay.csproj` Release: 0 errors, 0 warnings. Full `Arc.slnx` Release: 0 errors, 1 pre-existing `CS0436` in `TestApps/Chronicle` that reproduces on `main`.
- 884 specs passing (804 before, +80). Each spec was confirmed to fail against pre-fix code by reverting the change and re-running; the three items with no such spec (`SP0029` reservation, the nullable/optional specs, the options refactor) are called out as guards rather than regressions by construction.
- Determinism: two runs each over the real application and `TestApps/AspNetCore`, byte-identical.
- Against the real production application, the generated document now reads back under the official compiler with **0 errors and 0 warnings** (previously 1 warning on line 787).

**Diagnostic counts on that application, same commit lineage before and after:**

| Code | Before | After | |
|---|---|---|---|
| SP0021 | 1 | **0** | #2394 — the event resolves and is imported |
| SP0025 | 3 | **0** | #2395 — an `obj/` generator folder no longer counts as slice source |
| all others | unchanged | unchanged | |
| **total** | **1065** | **1061** | |

The whole document diff is the one `import` line plus its blank line, and 32 constraint names going from `uniquetimesheetstart` to `UniqueTimesheetStart`. Nothing else moved.

Note these differ from the figures circulating earlier (`SP0015 ×4`, `SP0020 ×9`, `SP0024 ×1`): those came from the published 20.65.0 through the CLI, not from current `main`. The before/after above are apples-to-apples from one code path.

**`Directory.Packages.props`** gained `Microsoft.CodeAnalysis.Workspaces.MSBuild`, `Microsoft.Build.Locator` and `Microsoft.NET.StringTools` for the harness. The harness sets `TargetFrameworks` (plural) to `net10.0` deliberately — the singular form leaves the repository's Release multi-targeting in place and restore then hits an NU1109 downgrade on `Microsoft.Extensions.Logging` for net8/net9.

woksin:
Superseded by #2408, which now targets `main` directly and carries all 19 commits from this branch, #2407 and `feat/screenplay-phase2` as one change with combined release notes. Nothing is lost — every commit here is an ancestor of that branch, and the branch is kept.

Closing this so there is one PR to review and one set of release notes rather than three that have to be merged in order.

PR #2407, opened 2026-07-28
Report a partly compiling source as a warning when recovery still stands

woksin:
## Changed


woksin:

**Stacked on #2406** — this targets `feat/screenplay-phase1`, not `main`, because it builds on that branch. It must merge after #2406; GitHub will retarget it automatically once that lands.

**The rule.** `SP0024` is an Error exactly when the number of artifacts recovered from a declaration that no compilation error sits inside is **zero**; otherwise it is a Warning and `IsSuccess` stays true. That single count collapses both cases the issue names — nothing recovered at all, and everything recovered coming out of a declaration an error is written inside.

A threshold ("near-zero", or a proportion) was deliberately rejected: the same recovery would pass for a large application and fail for a small one, and no cut-off is defensible. Zero is the only count that means recovery was *prevented* rather than dented. "Inside a declaration" is measured against every `DeclaringSyntaxReference`, so an error in a build-generated partial counts as inside the artifact.

**The suppression change is the part worth reviewing hardest.** `SP0024` previously suppressed `SP0034` (the generator reading its own document back) unconditionally. That is now narrowed to the Error branch only. As an Error, `SP0024` says nothing recovered can be trusted, so a poor document is the broken build rather than a second defect. As a Warning it says the opposite — the model stands — and suppressing `SP0034` there would hand back a document the Screenplay compiler rejects while reporting `IsSuccess == true`, which is the exact outcome `SP0034` was added to make impossible. Specified in both directions.

`SP0024 → SP0001` suppression is unchanged and is now provably vacuous in the Warning branch: zero artifacts ⟺ zero slices, so a Warning is only reachable when something was recovered, and `SP0001` could never fire there.

**A live demonstration of the issue, found by accident.** While verifying, the end-to-end gate reported `SP0024` with 41 errors on `TestApps/AspNetCore` — because that project had not been built in the worktree, so the workspace resolved nothing. Pre-fix that state exits 1; post-fix it exits 0 with a Warning. CI is green because it restores and downloads the build artifact, but it is worth noting that the gate is one artifact-download change away from meeting this same condition for real.

**Verification**

- `Screenplay.csproj` Release: 0 errors, 0 warnings. `Screenplay.Specs` and `Screenplay.EndToEnd` likewise.
- 928 specs passing (884 before, +44). 18 were confirmed to fail against pre-fix code, spanning both branches and the boundary; the two Error-branch severity facts pass either way by construction, and only their message assertions fail.
- Each commit verified standalone on a detached HEAD.
- End-to-end gate: exit 0, "The generated document reads back clean".
- The motivating case is proven hermetically from source strings: a slice that compiles plus a helper referencing a non-existent generated designer type yields 1 compilation error, 2 artifacts recovered, **Warning**, no error-severity diagnostic, and a document that compiles with `IsSuccess == true`.

**One imprecision, left deliberately.** The Warning's "N artifacts recovered" counts what the readers read. Where two queries in a slice collide irreconcilably, one is dropped and reported separately as `SP0022`, so the count can exceed the model's by that one. It is evidence in a message rather than a contract.

woksin:
Superseded by #2408, which now targets `main` directly and carries all 19 commits from #2406, this branch and `feat/screenplay-phase2` as one change with combined release notes. Nothing is lost — every commit here is an ancestor of that branch, and the branch is kept.

The SP0024 calibration and its narrowing of the SP0034 suppression are both described in the consolidated reviewer comment on #2408.

Issue #2403, opened 2026-07-28
Screenplay generation: recover FluentValidation property-reference operands and .When conditions


woksin:
## Summary

The validation chain reader drops two recoverable FluentValidation shapes:

1. **Property-reference operands**: `GreaterThanOrEqualTo(x => x.StartDate)` → `SP0011 [Warning] The 'GreaterThanOrEqual' rule on 'EndDate' carries no value to compare against and was left out`. The lambda names a sibling property — statically resolvable — and the Screenplay rule grammar already *parses* a bare path operand (`endDate >= startDate`; operands are mapping-source expressions), so the emission target exists today.
2. **`.When(...)` conditions**: rules guarded by a condition are dropped entirely. Simple condition lambdas (property comparisons against constants/enum members) are the same shapes the handler-body reader already recovers for `produces when`.

## Evidence (real application)

Generating from Ada produced 6 × SP0011 of shape (1) (`EndDate`, `ExpiresOn`, `When` properties) and multiple `.When(...)` drops inside SP0016 clusters. Combined with the whole-command `RuleFor(c => c)` business rules (18+, genuinely inexpressible today), the generated document's `validate` blocks retain only the string-length/regex layer — the domain-rule layer is absent.

## Suggested fix

1. When a comparison rule's operand is a lambda selecting a property of the validated type, emit the property path as the operand (`endDate >= startDate`). Note for the Screenplay side: the semantics of property-path operands should be stated in the grammar docs — tracked in Cratis/Screenplay#32.
2. When `.When(...)` carries a statically readable condition, hold the rule until the language grows rule-level conditions (Cratis/Screenplay#32) — until then, include the condition text in the SP0016 diagnostic so the drop is inspectable instead of opaque.
3. Whole-command state-based rules (`RuleFor(c => c).Must(...)` against injected read-model state) stay dropped until Cratis/Screenplay#32 lands a declarative form; the existing diagnostic is adequate there.

## Environment

- `Cratis.Arc.Screenplay` 20.65.0 (`Analysis/Validation/ValidationChainReader.cs`)
- Upstream language issue: Cratis/Screenplay#32

woksin:
## Resolved on `main`

Both shapes landed in #2418:

> A validation rule comparing one property against another is written as the comparison it is, rather than dropped as carrying no value (#2403)
> A validation rule guarded by a condition names that condition when it is left out, so what was dropped can be read (#2403)

[...]

Issue #2400, opened 2026-07-28
Screenplay generation: concepts referenced only from child records and read models are dropped, including PII-marked ones


woksin:
## Summary

Concepts are only emitted when referenced from positions the generator declares (event/command properties, query parameters and returns). Concepts that are referenced **only** from inside complex child records or read-model properties are dropped from the document — silently, with no diagnostic. In a real application this includes PII-marked concepts, so the generated document **understates the application's PII surface** — which cuts directly against Screenplay's "concepts carry compliance" design goal.

## Evidence (real application)

Diffing the generated Ada document (105 concepts) against the source (120 Core concept types, counting `ConceptAs<T>` and `EventSourceId<T>` derivatives) shows 15 concepts missing, none of which appear as a top-level event/command property type:

| Concept | PII | Where it lives |
|---|---|---|
| `FirstName` | **yes** | inside `PersonalDetails` child record |
| `LastName` | **yes** | inside `PersonalDetails` child record |
| `PhoneNumber` | **yes** | inside `PersonalDetails` child record |
| `AnswerText` | **yes** | inside `RequirementAnswer` child record (carried by events via `IEnumerable<RequirementAnswer>`) |
| `CriterionNote` | **yes** | inside evaluation child records |
| `EvaluationScore`, `SkillName`, `VoucherLineHours`, `WebsiteUrl`, `ActivityFeedKey`, `MagicLinkRecipientId`, `TripletexInvoiceId`, `TripletexPaymentId`, `TripletexCustomerPaymentId`, `RequestEmailParserConfigurationId` | no | child records / read-model-only positions |

All five PII concepts carry `[PII]` and validation rules in source (`FirstName: NotEmpty().MaximumLength(100)`, `PhoneNumber: Matches(@"^\+?[\d\s\-().]{7,20}$")`, …) — none of that reaches the document.

Related: the document *does* reference the child record types themselves in event property positions, e.g.

```
event TimesheetMonthApproved
  ...
  days ApprovedDayLine[]
```

`ApprovedDayLine` is never declared anywhere in the document — the Screenplay grammar has no complex-type declaration construct (companion language issue filed on Cratis/Screenplay). So today the document both references shapes it cannot define *and* loses the concepts inside them.

## Suggested fix

[...]

Issue #2397, opened 2026-07-28
Screenplay generation: recover validation messages from WithMessage(_ => Constant) lambdas (SP0016)


woksin:
## Summary

Validation-message recovery only accepts messages given as constant values directly. The very common FluentValidation shape `WithMessage(_ => SomeMessages.Constant)` — a lambda returning a constant — is rejected with `SP0016 [Warning] A message was declared without a constant value or without a rule preceding it, and was left out`. In an application that uses the message-constants pattern consistently, **every** validation message in the generated document is lost.

## Evidence (real application)

Generating from the Ada application produced **643 SP0016 warnings**, and the generated document's `validate` blocks carry no `message "…"` clauses at all. Ada's validators consistently use:

```csharp
RuleFor(c => c.Name)
    .NotEmpty()
    .WithMessage(_ => AccountsMessages.Name_Required);
```

where `AccountsMessages.Name_Required` is a `const string`. The Screenplay grammar has first-class support for messages (`not empty message "…"`), so the loss is purely on the recovery side.

Two compounding notes:

- The lambda body here is a plain member access to a `const` field — resolvable with the semantic model (`IFieldSymbol.HasConstantValue`), no evaluation needed.
- In Ada the message classes are *source generated*, so recovering them also requires the host to run source generators before analysis (see the companion SP0024 issue) — but the lambda shape fails even for hand-written const classes.

## Suggested fix

In the validation chain reader, when `WithMessage` receives a lambda whose body is a field/property reference with a compile-time constant value (or a literal), use that constant as the message. Keep SP0016 for genuinely computed messages (interpolation, method calls, localization lookups).

## Environment

- `Cratis.Arc.Screenplay` 20.65.0 (`Analysis/Validation/ValidationChainReader.cs`)

woksin:
Correction to the evidence in this issue, checked against the current `Ada` tree while implementing it (PR Cratis/Arc#2408).

**The fix described here is implemented, but it recovers nothing on this application — and the reason matters.**

All **510** `WithMessage` calls are the lambda form, as the issue says. But every one of them points at a resx-generated designer **property**:

```csharp
internal static string SomeMessage {
    get { return ResourceManager.GetString("SomeMessage", resourceCulture); }
}
```

[...]

Issue #2396, opened 2026-07-28
Screenplay generation: cross-slice query imports in screens are dropped silently


woksin:
## Summary

Screen analysis keeps only imports that name a query **the screen's own slice declares**; imports of queries declared in *other* slices of the same application are dropped — silently, with no diagnostic. Real Event-Modeling screens routinely aggregate several read models, so the generated documents systematically understate what screens actually bind.

## Evidence (real application)

In the Ada application, `Invoicing/Preparation/Queue/PreparationTab.tsx` imports four queries:

```ts
import { AllCustomers } from '../../../Customers/Listing/Listing';
import { All as AllBillableEngagements } from '../BillableEngagements/BillableEngagements';
import { All as AllDeviationCandidates } from '../Deviations/Deviations';
// plus its own slice's All (AdvisorInvoicePreparationQueue)
```

The generated document records only the same-slice binding:

```
screen PreparationTab
  file Invoicing/Preparation/Queue/PreparationTab.tsx
  data AdvisorInvoicePreparationQueue[] via query All
```

Across the whole application, only **5 of 180 screens** end up with any `data … via query` binding, largely because of this restriction (plus the ViewModel indirection pattern, where the query import lives in a sibling `*ViewModel.ts` rather than the `.tsx`).

`ScreenImports`' remark states the rationale: "an import naming something the slice does not declare says nothing and is dropped." That is true for arbitrary imports — but an import that resolves to a **query declared elsewhere in the same application model** is not "nothing"; it is exactly a data binding, and it is currently discarded without even a diagnostic, which contradicts the generator's own nothing-dropped-silently principle.

## Root cause is partly a language limitation

The Screenplay grammar's `data <Type> via query <Name>` takes a bare name. Generated documents contain many identically named queries (`All` appears 21 times in Ada — C# query method names are only unique per read model), so a cross-slice reference by bare name would be ambiguous. A companion issue on Cratis/Screenplay proposes defining reference scoping / qualified query references. 

## Suggested fix (in order of increasing ambition)

1. **Diagnose the drop today**: when a kept-out import matches a query declared in *another* slice of the model, emit an Information/Warning diagnostic ("screen X imports query Y declared in slice Z, which the Screenplay grammar cannot express as a binding, so it was left out"). This restores the honesty contract without waiting on the grammar.
2. When the grammar grows qualified references, emit cross-slice bindings.
[...]

PR #2408, opened 2026-07-28
Recover more of an application and report what cannot be recovered

woksin:
## Added

## Changed

## Fixed


woksin:

**This is agent-authored code that no human has read yet.** Green CI is not sign-off. **Do not merge without explicit review.** Stacked on #2407, which is stacked on #2406; it must merge last of the three.

## Three of the issues' premises turned out to be wrong

This is the most important part of this PR. The acceptance criteria could not be met as written, because the evidence they rest on does not hold against the current source.

**#2397 — the message recovery does not fire on the real application, and the 643 figure is not what it looks like.** All 510 `WithMessage` calls there are the lambda form, but every one points at a **resx-generated designer property** — `internal static string X => ResourceManager.GetString(…)` — which is a runtime, culture-dependent lookup, not a constant. `HasConstantValue` cannot reach it, and "keep SP0016 for genuinely computed messages" applies. `SP0016` therefore stays at 643. The 643 is also not all message drops: the real split is 268 `RuleFor(c => c)` whole-command rules, **247** messages, 59 `Must`, 28 `NotEqual`, 20 `MustAsync`, 12 `When`, 6 `InclusiveBetween`, 3 `IsInEnum`. Recovering those messages needs `$strings.` localization-key emission from the resx key — a different change, not attempted here. What did improve: those 247 now name the expression they could not write down, and the 12 `When` drops name their condition.

**#2403 — 6 → 2, not 6 → 0.** Only four of the six `SP0011`s were the property-reference shape the issue describes. The other two are `.GreaterThan(businessClock.Today())` (a method call) and `.GreaterThan(DateOnly.MinValue)` (`static readonly`, not a constant); both correctly stay dropped.

**#2400 — only 7 of the 15 named concepts are recoverable, and the three PII ones are not among them.** `FirstName`, `LastName`, `PhoneNumber`, `WebsiteUrl`, `MagicLinkRecipientId` and the three Tripletex ids are **not referenced by any artifact** in the current tree. `FirstName` and `LastName` appear nowhere outside their own declaration files; `PhoneNumber` appears only in an interface method signature and its implementation; `WebsiteUrl` only in specs; the Tripletex ids only as handler-returned event source ids, already reported as `SP0013`. That application's `PersonalDetails` carries `FullName`/`DisplayName`/`Bio`, not the three names the issue's table lists. Declaring an unreferenced concept would contradict the registry's "declare only what is referenced" rule, so it was not done. **This was independently re-checked against the source before writing this.**

Personal data did improve where it was reachable: `@pii` went 27 → 29, with `AnswerText` and `CriterionNote` both emitting as `String @pii` with their `validate` blocks intact. Concepts overall went 105 → 130.

**#2396 — the diagnostic works; the optional ViewModel hop is inert here.** `SP0036` reports 29 cross-slice bindings, resolving each import through its module path so the correct slice is named; one screen reports five. The one-hop follow through sibling `*ViewModel.ts` files is implemented and specified but adds **0** bindings on that application — its 17 ViewModels import commands, prop types and helpers, never a query proxy; the query imports sit in the `.tsx` directly. Which query a screen binds when a bare name collides across slices was deliberately left alone, as changing it would remove bindings.

## Diagnostics, before and after, same commit lineage

| Code | Before | After | |
|---|---|---|---|
| SP0011 | 6 | **2** | four property-reference operands now emitted |
| SP0016 | 643 | 643 | text improved, count unchanged — see above |
| SP0031 | 0 | **1** | genuine and newly reachable: two `TimelineEntryKind` enums were both unreachable before; now both are reached and the collision is reported, first wins deterministically |
| SP0035 | 0 | **13** | new — one per record an event or command carries whose shape is undeclarable (upstream Cratis/Screenplay#29) |
| SP0036 | 0 | **29** | new — cross-slice query bindings previously dropped in silence |
| all others | unchanged | unchanged | |
| **Total** | **1061** | **1100** | |

## Verification

- `Screenplay.csproj` Release after `dotnet clean`: 0 errors, 0 warnings.
- 973 specs passing (928 before, +45); the report lists exactly which of them were confirmed to fail against pre-fix code, and it is a subset — the rest are perimeter guards that cannot fail by construction.
- End-to-end gate on `TestApps/AspNetCore`: exit 0, reads back clean, diagnostics unchanged.
- Real application: exit 0, 5530 lines, reads back clean under the Screenplay compiler.
- Determinism: two runs, document and diagnostics both byte-identical.

## Caveat

`TypeRegistry.cs` is now **251 lines**, over the repository's ~200-line guideline (244 before). A split into a separate concept registry is available; it was left rather than churn the area for size alone. Worth a reviewer's opinion.

woksin:
Consolidated reviewer context for the whole change — deliberately kept out of the description. This PR now supersedes #2406 and #2407; all 19 commits are here.

## Where I would look first

**The suppression interaction between #2393 and the document check.** `SP0024` previously suppressed `SP0034` (the generator reading its own printed document back) unconditionally. That is now narrowed to the Error branch. As an Error, `SP0024` says nothing recovered can be trusted, so a poor document is the broken build rather than a second defect. As a Warning it says the opposite — and suppressing `SP0034` there would hand back a document the Screenplay compiler *rejects* while reporting `IsSuccess == true`, the one outcome `SP0034` exists to make impossible. Two changes interacting in a way neither issue anticipated.

**The `SP0024` rule itself.** Error exactly when the count of artifacts recovered from a declaration no compilation error sits inside is **zero**; Warning otherwise. A threshold was deliberately rejected — the same recovery would pass for a large application and fail for a small one. "Inside a declaration" is measured against every `DeclaringSyntaxReference`, so an error in a build-generated partial counts as inside the artifact.

**The end-to-end CI gate** (`Source/DotNET/Screenplay.EndToEnd/`, wired into `dotnet-build.yml`) is the most valuable thing here and is not in the release notes. It generates from `TestApps/AspNetCore` and compiles the result with the `Cratis.Screenplay` compiler, failing on **any** compiler diagnostic including warnings. Two of the defects found before this package first shipped were invisible to hermetic specs by construction.

Worth knowing: because the workflow triggers on `pull_request_target`, a job added in a PR runs from the **base** branch and cannot execute on its own PR. I proved the gate works by dispatching the workflow against the branch directly — it generated 53 lines and reported "reads back clean" on a runner. Its first `pull_request_target` run will be after this merges.

## Three of the issues' premises did not hold

Checked against the source, not taken from the issues:

- **#2397** — all 510 `WithMessage` calls in the reference application are the lambda form, but every one points at a resx-generated designer *property* (a runtime, culture-dependent lookup), not a constant, so `HasConstantValue` cannot reach it. `SP0016` stays at 643. That 643 is also not all message drops: it is 268 whole-command `RuleFor(c => c)` rules, 247 messages, 59 `Must`, 28 `NotEqual`, 20 `MustAsync`, 12 `When`, 6 `InclusiveBetween`, 3 `IsInEnum`. Recovering those messages needs `$strings.` localization-key emission — a different change, not attempted.
- **#2403** — 6 → 2, not 6 → 0. Only four were the property-reference shape; the others are a method call and a `static readonly`, correctly still dropped.
- **#2400** — only 7 of the 15 named concepts are recoverable. `FirstName`, `LastName`, `PhoneNumber`, `WebsiteUrl`, `MagicLinkRecipientId` and three Tripletex ids are **not referenced by any artifact** in the current tree: the first two appear nowhere outside their own declaration files, `PhoneNumber` only in an interface method signature and its implementation. Declaring an unreferenced concept would contradict the registry's "declare only what is referenced" rule. Personal data did improve where reachable — `@pii` 27 → 29, concepts 105 → 130.

## Diagnostics on the reference application, across the whole change

| Code | Before | After | |
|---|---|---|---|
| SP0011 | 6 | **2** | property-reference operands now emitted |
| SP0016 | 643 | 643 | text improved, count unchanged — see above |
| SP0021 | 1 | **0** | the external event is imported |
| SP0024 | 1 | **0** | resolved upstream in the host; calibration makes it a warning regardless |
| SP0025 | 3 | **0** | an `obj/` generator folder no longer counts as slice source |
| SP0031 | 0 | **1** | genuine and newly reachable: two same-named enums were both unreachable before |
| SP0035 | 0 | **13** | new — records whose shape is undeclarable (upstream Cratis/Screenplay#29) |
| SP0036 | 0 | **29** | new — cross-slice query bindings previously dropped in silence |

## Verification

- `Screenplay.csproj` Release after `dotnet clean`: 0 errors, 0 warnings.
- **973 specs** passing, up from 804 when this line of work started. Each phase's new specs were confirmed to fail against pre-fix code by reverting and re-running; the reports name exactly which, and which are perimeter guards that cannot fail by construction.
- End-to-end gate: exit 0, "reads back clean", both locally and on a CI runner.
- Reference application: 5530 lines, reads back clean under the official Screenplay compiler, byte-identical across two runs.

## Caveats

- `Analysis/Types/TypeRegistry.cs` is **251 lines**, over the ~200-line guideline. A split into a separate concept registry is available; left rather than churn the area for size alone.
- The gate does not carry the resx `PrepareResources` workaround the CLI has, so it would trip on an application whose resource classes are build-generated. `TestApps/AspNetCore` has none, so CI is unaffected.
- #2398's "prefer the declaring type name minus a `Constraint` suffix" was **not** adopted: one `IConstraint` may declare several rules in a single `Define`, and naming them all after the type collapses them into duplicate `constraint` names in a slice.
- #2396's optional one-hop follow through sibling `*ViewModel.ts` files is implemented and specified but adds nothing on the reference application, whose view models import commands and prop types rather than query proxies.

woksin:
Repository-wide verification, which should have run before this was pushed rather than after.

- `dotnet clean Arc.slnx -c Release` then `dotnet build Arc.slnx -c Release`: **0 errors, 1 warning** — the pre-existing `CS0436` in `TestApps/Chronicle`, which reproduces on `main` without this branch.
- Full suite, `dotnet test Arc.slnx -c Release`: every project passes. Two runs reported a single failure each in `Arc.Core.Specs` on net9.0 and net10.0; re-running that project in isolation gives **1552/1552 on both**, so it is the known parallel timing flake in that suite rather than anything from this change.
- `Screenplay.Specs`: 973 across net8.0/net9.0/net10.0.
- Working tree clean afterwards, including `TestApps` — `Commands.ts` did not regenerate dirty.

woksin:
Three follow-ups pushed (`a8422af0`, `0569cb8c`). None changes behaviour, so the release notes are unchanged.

**Diagnostics now name the upstream language issue they wait on.** `SP0013` → Cratis/Screenplay#33, `SP0015` → #30, `SP0016` → #32, `SP0020` → #39, written into the explanation rather than appended as a tag.

Five candidates were considered and **rejected**, which is the more useful half:

- `SP0011` → #32 — it fires only when neither a sibling-property lambda nor a compile-time constant could be read. Cross-property comparison is *supported* on this branch, so what remains is a source-reading limit, not a language gap; resolving #32 would not silence it. Linking would tell a reader it waits on the language when it mostly waits on us.
- `SP0019` → #36/#37 — fires for transport-level returns (`ActionResult`) and type-parameter returns; neither is identity scoping nor observability.
- `SP0027` → #35 — a `produces when` guard on aggregate state is a different mechanism from a command declaring what it reads.
- `SP0012`, `SP0014`, `SP0017`, `SP0018`, `SP0028` — nothing in the upstream list covers them.

Also verified as having no corresponding diagnostic at all: #31 (a CLI concern), #34 (no side-effect concept exists — a returning reactor is silently recorded as translating), #35, #36, #37 (`ISubject<T>` is silently stripped as a wrapper), #38 (`IsPii` is a bare bool with no reason field). Several of those silences may deserve diagnostics of their own; not in scope here.

**`TypeRegistry` split**, 251 → **160**, with a new `ConceptRegistry` at **163** — both under the guideline. The seam is real rather than a cut: the type was answering *what name do I write* and *what are the application's concepts* at once, and the second has its own rules about what counts, which declaration wins on a name collision, and how a later `@pii` mark finds its concept. A duplicated enum/`ConceptAs` branch pair collapsed out of it. Public surface unchanged.

**Comparison against the reference document.** Fresh output 5,526 lines against the 5,348-line reference: 52 removed, 230 added, 61 hunks, and **every line classified** — nothing unplaced.

- *Intended fixes* (188 added, 31 removed): 24 recovered concept declarations including two carrying `@pii`; 31 constraint names becoming readable; the one `import` for the externally declared event; 4 recovered comparisons.
- *Source drift in the application itself* (42 added, 21 removed): a new `EmailReservations` slice with its event, and two reducers replaced by real projections — the C# type names changed, which only the application can do.

One name looked wrong and was chased down rather than waved past: `UniqueTimesheetMonthendReminder`. The application's own source declares `"unique-timesheet-monthend-reminder"`; the generator reproduced the misspelling faithfully.

**Determinism and refactor safety in one check:** the document generated before the split, after it, and on a second run afterwards all hash `d142fcca…`.

**Caveat on that comparison:** the reference application's working tree has 87 uncommitted modifications and was being edited concurrently during the run, so the drift classification is against a live tree rather than a commit. It does not weaken the byte-identity guard above, and `/Users/sindrewilting/Ada.play` was not modified.

Verification: Release 0 errors / 0 warnings; **973** specs, count unchanged as a refactor requires; end-to-end gate exit 0, reads back clean; worktree clean.

woksin:

Issue #2399, opened 2026-07-28
Screenplay generation: multi-project applications are only partially described


woksin:
## Summary

`ScreenplayGenerator.Generate` takes a single `Compilation`, and `ArtifactCatalog.From` walks only `compilation.Assembly.GlobalNamespace`. A multi-project Arc application (Domain/Application/Api layering, or a main app plus sibling contexts) is therefore only partially described: artifacts in referenced projects are invisible, and cross-project event references degrade to `SP0021` warnings.

## Evidence (real application)

Ada's solution has several projects containing Arc artifacts (`Core`, `Lobby`, `Lobby.Contracts`, `Integrations`, …). Generating from `Core.csproj` yields a document that:

- describes only `Core`'s artifacts,
- refers to `InvitationToJoinAdaAccepted` (declared in a referenced project) that it never introduces (`SP0021`; the official Screenplay compiler then warns on the generated file).

The CLI integration (branch `feature/arc-screenplay` on Cratis/cli) *narrows* a solution to exactly one project and errors when two candidates remain — so pointing the tool at a layered solution either fails or silently describes a fraction of the application.

## Suggested fix

Add an overload accepting multiple compilations and merge the catalogs:

```csharp
ScreenplayGenerationResult Generate(IReadOnlyList<Compilation> compilations, ScreenplayOptions options);
```

- Walk each compilation's own assembly; union the artifact catalogs (ordered by assembly name, then the existing deterministic ordering, so output stays reproducible).
- Slice/namespace attribution already works per-symbol, so features from different projects land in the right modules; `SegmentsToSkip`/module inference may need a per-assembly default.
- The CLI can then hand over every project of the solution (minus specs/tests) instead of erroring on ambiguity.

This is the difference between "describes the project you pointed at" and "describes the application" — the stated goal of the tool.

## Environment

- `Cratis.Arc.Screenplay` 20.65.0 (`Analysis/ArtifactCatalog.cs`, `IScreenplayGenerator.cs`)

woksin:
## Resolved on `main`

Landed in #2418:

> An application split across several projects is described as one document: a namespace two projects declare into is one slice, a concept is declared once however many projects refer to it, and an event a sibling project declares is declared rather than imported (#2399)
> A command handing work to an aggregate root declared in another project is read instead of failing (#2399)

The overload the issue proposed is on the interface verbatim:

```csharp
[...]

Issue #2409, opened 2026-07-28
Screenplay generation: emit $strings. keys for validation messages that resolve to resources


woksin:
Validation messages that resolve to a resource string are currently dropped with `SP0016`, and cannot be recovered by reading a constant.

### Why the constant route does not work

A localized application writes its messages as:

```csharp
RuleFor(_ => _.Title).NotEmpty().WithMessage(_ => RequestMessages.TitleRequired);
```

where `RequestMessages.TitleRequired` is generated from a `.resx` by the `MSBuild:Compile` generator:

```csharp
internal static string TitleRequired {
    get { return ResourceManager.GetString("TitleRequired", resourceCulture); }
}
```

That is a runtime, culture-dependent lookup. It has no constant value, so `HasConstantValue` / `GetConstantValue` return nothing and the message is correctly reported as computed. #2397 recovers the lambda-returning-a-literal and lambda-returning-a-`const` shapes, but those do not occur in a localized application.

Measured on a real production application: **510** `WithMessage` calls, **all** of this shape, accounting for **247** of its `SP0016` diagnostics.

### Suggested direction

The Screenplay language has a `$strings.` form for exactly this. Rather than resolving the message to text at generation time — which would bake one culture into a document meant to describe the application — emit the **key**:

```
message $strings.TitleRequired
```

The key is statically available: it is the `GetString` argument in the generated designer property, and equivalently the property name. Recovering it means recognizing that a symbol is a resx designer property and taking its name, rather than trying to evaluate it.

Points worth deciding:

- Whether to qualify the key with the resource class (`$strings.RequestMessages.TitleRequired`) or emit it bare, and what the language's `$strings.` form actually accepts.
- Whether to verify the key exists in the `.resx` at generation time, or trust the symbol.
- What to do when a message is composed rather than referenced directly (`string.Format(...)`, interpolation) — those should stay `SP0016`.

### Acceptance

A validation rule whose message references a resource designer property emits a `$strings.` message rather than being dropped; `SP0016` on the reference application falls by roughly 247; the document still compiles under the Screenplay compiler; generation stays deterministic. Messages that are genuinely computed continue to be reported.

Related: #2397 (where this was discovered and why the constant route was insufficient).

woksin:
Correction to my own estimate in this issue, from implementing it (PR #2413).

**I predicted `SP0016` would fall by ~247. It falls by 133** — 643 → 510.

[...]

Issue #2404, opened 2026-07-28
Screenplay generation: recover 'specification' blocks from Chronicle integration specs


woksin:
## Summary

The Screenplay language has a first-class `specification` construct (given events / given readmodel / when command / then event / then error), but the generator never emits specifications — they were classified as uninferable in PR #2384. For Chronicle **integration specs** written in the project's own house style, that classification is too pessimistic: the specs follow a rigid, statically readable pattern that maps 1:1 onto the construct.

## Evidence (real application)

Ada contains ~1,450 spec files co-located with their slices (`when_*/and_*.cs`), and the integration specs among them are exactly the documented Cratis shape:

```csharp
public class and_a_second_start_for_the_same_timesheet_is_appended(context context) : Given<context>(context)
{
    public class context(ChronicleOutOfProcessFixture fixture) : given.an_http_client(fixture)
    {
        async Task Establish() =>
            await EventStore.EventLog.Append(id, new TimesheetStarted(...));   // → given TimesheetStarted

        async Task Because() =>
            Result = await Client.ExecuteCommand(..., new StartMonth(...));    // → when StartMonth
    }

    [Fact] void should_fail() => _result.ShouldHaveConstraintViolationFor(UniqueTimesheetStartConstraint.Name);  // → then error
}
```

- `Establish` appends via `EventStore.EventLog.Append(id, new SomeEvent(...))` → `given SomeEvent` with property values from the object-creation arguments (constants recoverable the same way produces-mappings already are).
- `Because` executes exactly one command via `ExecuteCommand` → `when SomeCommand` with its values.
- `[Fact]` assertions: `ShouldHaveAppendedEvent<T>` → `then T`; `IsSuccess.ShouldBeFalse()` / `ShouldHaveConstraintViolationFor` → `then error`; read-model assertions → `then readmodel`.

The Screenplay grammar's `specification` blocks carry precisely these parts and nothing more, so partial recovery is honest by construction: recover the specs whose Establish/Because/assert steps are all statically readable, and diagnose the rest ("this spec uses computed values / helper indirection, so it was left out") per the existing nothing-dropped-silently discipline.

## Why it matters

Specifications are the behavioral contract layer of a Screenplay — a document that states what slices *do* but carries none of the scenarios proving it loses half the story Stage/Studio could replay. For applications like Ada the scenarios already exist in source, in a discoverable convention, next to the slices they describe.

## Suggested scope

[...]

PR #2410, opened 2026-07-28
Describe an application split across several projects as one document

woksin:
## Added

## Fixed

### DIFF
--- a/Source/DotNET/Screenplay.EndToEnd/Program.cs
+++ b/Source/DotNET/Screenplay.EndToEnd/Program.cs
@@ -8,7 +8,7 @@ using Cratis.Screenplay;
 if (args.Length < 2)
 {
     Console.WriteLine("Usage:");
-    Console.WriteLine("  Cratis.Arc.Screenplay.EndToEnd <project> <output-file>");
+    Console.WriteLine("  Cratis.Arc.Screenplay.EndToEnd <project-or-solution> <output-file>");
 
     return 2;
 }
@@ -19,21 +19,26 @@ var output = Path.GetFullPath(args[1]);
 Console.WriteLine($"Generating the Screenplay document of '{project}'");
 
 var failures = new List<string>();
-var compilation = await ProjectCompilation.Of(project, failures);
+var compilations = await ProjectCompilation.Of(project, failures);
 
 foreach (var failure in failures)
 {
     Console.WriteLine($"  workspace: {failure}");
 }
 
-if (compilation is null)
+if (compilations.Count == 0)
 {
     Console.WriteLine($"'{project}' yielded no compilation, so there is nothing to generate from");
 
     return 1;
 }
 
-var generated = new ScreenplayGenerator().Generate(compilation, new ScreenplayOptions());
+foreach (var loaded in compilations)
+{
+    Console.WriteLine($"  project: {loaded.AssemblyName}");
+}
+
+var generated = new ScreenplayGenerator().Generate(compilations, new ScreenplayOptions());
 
 Directory.CreateDirectory(Path.GetDirectoryName(output)!);
 await File.WriteAllTextAsync(output, generated.Source);
--- a/Source/DotNET/Screenplay.EndToEnd/ProjectCompilation.cs
+++ b/Source/DotNET/Screenplay.EndToEnd/ProjectCompilation.cs
@@ -9,53 +9,122 @@ using Microsoft.CodeAnalysis.MSBuild;
 namespace Cratis.Arc.Screenplay.EndToEnd;
 
 /// <summary>
-/// Reads a real project file into the compilation the generator is handed.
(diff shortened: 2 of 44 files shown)


woksin:

**This is agent-authored code that no human has read yet.** Green CI is not sign-off. **Do not merge without explicit review.** Stacked on #2408; it must merge after.

## API

```csharp
ScreenplayGenerationResult Generate(IReadOnlyList<Compilation> compilations, ScreenplayOptions options);
ApplicationModelAnalysis Analyze(IReadOnlyList<Compilation> compilations, ScreenplayOptions options);
```

The single-`Compilation` overloads keep their signatures and delegate to a list of one, so there is a single code path rather than two that can drift. Published surface unchanged.

**Ordering.** Compilations are sorted by assembly name, tie-broken by the ordinally-first source file path, *before* anything is read — caller order never reaches the model. What projects genuinely share is unified (one type registry, one aggregate-root catalogue, one cross-slice query set, one diagnostics sequence); everything reading symbols of a single compilation stays per project.

## The four judgment calls

- **A namespace two projects declare into is one slice**, not two. A contracts project publishing a slice's events beside the project handling its commands is the ordinary layered shape, and two slices of one name would be a document saying the same word twice. Name collisions within the joined slice keep the first and report **SP0037**.
- **`SegmentsToSkip` stays one value for the whole application.** A per-assembly value would need a dictionary keyed on assembly names that a caller configuring it has no reason to know, and the projects of one application share a namespace root.
- **Path roots** are resolved per project and then written relative to the directory the projects share, so paths read `Core/Features/…` and `Lobby/…` rather than two files both printing `Features/…`. For a single project this is identical to today. Where projects share nothing but the file-system root, each keeps its own root — never absolute — and that is reported as **SP0038**. Path roots were a defect once before, so this was checked directly: every `file` line on the real solution is relative.
- **Screens** stay per slice, but cross-slice query detection is now application-wide, so a screen reading through a query a *sibling project* declares is reported rather than dropped.

## A latent bug found on the way

A command handing work to an aggregate root in another project crashed with `ArgumentException: SyntaxTree is not part of the compilation`. This was **already reachable on the single-project path**, because a workspace loads project references as compilations. A new `SemanticModels` resolves a tree through whichever project wrote it; a tree in no supplied compilation is reported rather than silently dropped.

## Verification

- Release build: 0 errors, 0 warnings. All five commits were exported and built independently: 7 projects, 0/0 each.
- **1049 specs** (973 before, +76). Because every new spec needs the new API, load-bearingness was verified by reverting each piece of logic *individually* — the report lists which specs fail for each revert, covering slice union, shared source root, shared type registry, cross-project aggregates, cross-project policies, and compilation ordering.
- Two-project hermetic fixture with a real `CompilationReference`: one document, zero Screenplay-compiler diagnostics, no `SP0021`, the shared concept declared once. A second fixture proves both orderings give byte-identical text.
- End-to-end gate: exit 0, reads back clean, diagnostics unchanged.
- **Real solution, 6 non-spec projects:** exit 0, 5,639 lines, reads back clean. The `import Lobby.Invitations.InvitationToJoinAdaAccepted` present in the single-project document is **gone**, replaced by a real `feature Lobby` with `Publishing`/`Receiving` slices declaring the event — which is the point of the change.

## Two honest caveats

**The single-project regression check is parent-vs-mine under identical environment, not against the published baseline.** The first run reproduced the stated 5530 lines / 1100 diagnostics exactly; every run afterwards — *including runs of the unmodified parent code*, confirmed by stashing — gives 5527 / 1108, because that application's design-time build mutated its own gitignored `obj/` on first load. Byte-identity between parent and this branch was verified twice with `cmp` under the same state.

**`ScreenplayProjectSelection` in the CLI is not done.** Handing all non-spec projects of a solution to the generator still errors on ambiguity; that lives in Cratis/cli with an open PR against it and was out of scope here. The harness change in the last commit teaches the *test harness* to accept a solution — it is not the CLI and does not substitute for that work.

woksin:
Independent verification of the multi-project result, run separately from the work itself.

Generated from the reference application's **full solution** (`Ada.slnx`) with this branch: exit 0, **5,641 lines**, "The generated document reads back clean" under the Screenplay compiler.

- **0** `file` lines that are absolute or point into `obj/` — path roots were a real defect once, and they hold under the harder multi-project case.
- Paths root at `Core/` (224) and `Lobby/` (5), correctly disambiguated per project rather than both printing `Features/…`.
- **0** `import` lines — the dangling `InvitationToJoinAdaAccepted` import present in the single-project document is gone.
- `feature Lobby` is present, with the event declared by a real slice rather than referenced from outside.

Line count differs slightly from the figure in the report above (5,641 vs 5,639) because that application's working tree is being edited concurrently; it is drift in the source, not in the generator.

CI on this PR is fully green: 43 passing, 0 failing, 45 checks.

woksin:

PR #2411, opened 2026-07-29
Recover the specifications a slice carries into the document

woksin:
## Added

### DIFF
--- a/Source/DotNET/Screenplay.Specs/IntegrationTesting.cs
+++ b/Source/DotNET/Screenplay.Specs/IntegrationTesting.cs
@@ -1,0 +1,111 @@
+// Copyright (c) Cratis. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Cratis.Arc.Screenplay;
+
+/// <summary>
+/// Declares the testing surface a Chronicle integration specification is written against, as source.
+/// </summary>
+/// <remarks>
+/// The generator recognizes these by the fully qualified name of the type declaring each member, which is exactly
+/// what lets it read a specification without the testing packages being referenced. Declaring them here rather than
+/// referencing them keeps every specification about reading them hermetic, and asks the recognition the only
+/// question worth asking of it - whether the names alone are enough.
+/// </remarks>
+public static class IntegrationTesting
+{
+    /// <summary>
+    /// The path the surface is compiled as.
+    /// </summary>
+    public const string Path = "Library/Testing/IntegrationTesting.cs";
+
+    /// <summary>
+    /// The source of the surface.
+    /// </summary>
+    public const string Source = """
+        using System.Net.Http;
+        using System.Threading.Tasks;
+        using Cratis.Chronicle.Events;
+        using Cratis.Chronicle.EventSequences;
+
+        namespace Cratis.Arc.Testing.Commands
+        {
+            public class CommandScenario<TCommand>
+            {
+                public Cratis.Arc.Chronicle.Testing.Commands.CommandScenarioChronicleGivenBuilder<TCommand> Given => new();
+
+                public IEventSequence EventSequence => null!;
+
+                public Task<Result> Execute(TCommand command) => Task.FromResult(new Result());
+
+                public Task<Result> Validate(TCommand command) => Task.FromResult(new Result());
+            }
+
+            public class Result
+            {
+                public bool IsSuccess => true;
+            }
+        }
+
+        namespace Cratis.Arc.Chronicle.Testing.Commands
+        {
+            public class CommandScenarioChronicleGivenBuilder<TCommand>
+            {
+                public CommandScenarioSourceGivenBuilder<TCommand> ForEventSource(EventSourceId eventSourceId) => new();
+            }
+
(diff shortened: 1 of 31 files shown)


woksin:

**This is agent-authored code that no human has read yet.** Green CI is not sign-off. **Do not merge without explicit review.** Stacked on #2410, which is stacked on #2408; this merges last of the three.

## What is recognized

A class is treated as a Chronicle integration specification when it is a non-abstract top-level class with a `Because`, and either holds a `CommandScenario<T>` or touches the recognized testing surface. Matching is on fully-qualified metadata names, so neither testing package has to be referenced.

| Screenplay | in-process shape | running-host shape |
|---|---|---|
| `given E` | `Scenario.Given.ForEventSource(x).Events(new E(…))` | `EventLog.Append(x, new E(…))` |
| `given readmodel R` | `…ForEventSource(x).ReadModel(new R(…))` | — |
| `when C` | `Scenario.Execute/Validate(new C(…))` | `Client.ExecuteCommand(url, new C(…))` |
| `then E` | `ShouldHaveAppendedEvent<…, E>` | same |
| `then error "…"` | `ShouldHaveConstraintViolationFor(x)`, else `ShouldNotBeSuccessful`, `ShouldHaveValidationErrors`, … | same |

**The issue's evidence was out of date about its own reference application.** #2404 describes `EventLog.Append` and `ExecuteCommand`; that application uses `Scenario.Execute` **exclusively** and has zero occurrences of either. Supporting only the documented shape would have recovered nothing. Both are supported.

## Two judgement calls

**`then error ""` when the source names no reason.** Screenplay requires a string, and `ShouldNotBeSuccessful()` names none. Writing "rejected" or "validation" would be an invented sentence; the scenario's *name* already carries the reason. All 347 `then error` lines on the reference application are unnamed. If you would rather such scenarios were skipped than written with an empty reason, that is a one-line change.

**A conditional step takes the whole scenario with it.** A real correctness bug was found and fixed mid-implementation: `if (CandidatePersonErased) { …ReadModel(new ErasedPerson(…)); }` was read as an *unconditional* given for every derived spec, stating a world nobody specified. Anything under `if`/`switch`/loop/ternary/`&&`/`??`/lambda/`catch` now makes the scenario unreadable. Cost: 29 fewer specifications, 54 skips — a document that says less and is true.

## Placement, which the merge changed

Specifications are read **per project** (all semantic work on one compilation), but placed **after** `SliceUnion` joins the slices. `SpecificationPlacement.SliceOf` walks strictly upward for a namespace declaring a slice, so a scenario written in the project holding the handlers, against a slice whose events come from a contracts project, would find nothing above it and be dropped. Merging goes through `SliceUnion.Specifications`, reusing the union's existing "first of a name kept, the rest reported" rule.

## Volume, which was the open question

**+3,460 lines for 507 scenarios — about 7 lines each, 62% growth**, not the multiple I expected. Most values are runtime identities and get dropped, so a scenario averages 2.4 value lines. No option to disable was added, and I agree with that: if a larger application does make a document unwieldy, the right lever is a per-slice or per-module filter driven by a real complaint, not a global off switch added on speculation.

## Verification

- Release build 0 errors / 0 warnings across net8.0/net9.0/net10.0.
- **1106 specs**, exactly the 1030 + 1049 − 973 the merge predicted.
- End-to-end gate: exit 0, reads back clean.
- Reference application, single project: 9,110 lines, **507 specification blocks**, reads back clean.
- Reference application, **whole solution**: 9,290 lines, **520 specification blocks**, **0 `import` lines**, 229 `file` lines all relative (224 `Core/`, 5 `Lobby/`, 0 absolute) — this is the check that proves specifications and multi-project coexist.
- Determinism: byte-identical across consecutive runs for all three targets.

## Two corrections the agent made to its own work, worth reading

**A "zero specifications recovered" result was not the merge.** That application's `obj/project.assets.json` had been restored without the Debug-only test packages, so the compilation carried all 1,422 spec syntax trees but no xunit or testing package — 13,000+ errors and nothing matched. Proven pre-existing by building the *pre-merge* tip and getting byte-identical output with the same zero. A plain `dotnet restore` fixed it; no tracked file was touched.

**A "nondeterminism" finding was a broken guard, not a real defect.** Five distinct hashes across six runs looked like nondeterminism in the multi-project path. The cause was the application being edited concurrently, while the guard — `find -newermt "-30 minutes"` — silently matches nothing on BSD `find` and so reported "no concurrent edits" while files were being written. Redone with `touch`-marker files, everything is byte-identical.

## Known limits

- **Values held in fields are not followed one hop**, losing 77 scenarios to `.ReadModel(TargetUser)` / `Execute(_command)` built in `Establish`. Closable, and the highest-value follow-up.
- `then readmodel` is never emitted — nothing in the recognized assertion surface asserts read-model state, so there is no source shape to read it from, despite the issue listing it.
- `EventScenario`-based constraint specs are treated as unit-level and not reported; they append an event as their action, which `specification` has no `when` for.
- The solution load pulls in a netstandard2.0 analyzer project as an application project (`ProjectCompilation.IsSpecifications` filters by name only). It contributes nothing to the document, but the workspace warns about it. Pre-existing, untouched.

woksin:

PR #2413, opened 2026-07-29
Recover more of an application, describe it across projects, and report what cannot be recovered

woksin:
## Added

- A validation message looked up from a resource is written as the key it is looked up by, so the document states which message a rule carries without baking one culture into it (#2409)

## Changed

## Fixed


woksin:

**This is agent-authored code that no human has read yet.** Green CI is not sign-off. **Do not merge without explicit review.** Top of the stack: this → #2411 → #2410 → #2408 → `main`.

## The grammar was checked before anything was built on it

`$strings.` **is** accepted where a validation message goes, confirmed two independent ways: the grammar (`LocalizableString = StringLiteral | "$strings.", Path`) and the real parser (`ValidationRuleParser`, which accepts dotted `\w` segments), plus Screenplay's own round-trip spec for it. Arc's *emission* side was already wired for this and had simply never produced a key — all the work was in analysis.

## Three decisions

**Qualified, `$strings.<ResourceClass>.<Key>`** — not a style preference. In the reference application **6 keys collide across two resource classes** (`Approve_OrgNumberRequired`, `Submit_NotAdministered`, … in both `CustomersMessages` and `NetworkMessages`), carrying different text in different features. Bare keys would collapse those into one key that can hold only one string.

**The key comes from the `GetString` literal, not the property name.** `StronglyTypedResourceBuilder` mangles non-identifier characters when it makes the property name, so the literal is the real resx key and the property name is a derived approximation. Recognition is by the *shape* of the getter — a static string property whose getter contains exactly one `ResourceManager.GetString(...)` with a literal first argument — never by file name or naming convention. It deliberately misses a resource class compiled into a referenced assembly, where there is no getter syntax to read; guessing the key from the property name there was refused.

**No new diagnostic codes.** SP0041–SP0043 were allocated and went unused: every remaining drop is still "a validator declares a rule that could not be expressed declaratively", which is what `SP0016` already says, and its existing sentence already covers the resource case. A code a consumer cannot act on differently is not worth minting.

## The issue's estimate was wrong, and the correction is the interesting part

**#2409 predicted `SP0016` would fall by ~247. It fell by 133** — 643 → 510. That estimate was mine, and it conflated two different things.

The premise held: **509 of 510** `WithMessage` calls are the resource shape, none are composed, and all 510 are now readable. But recovering a message only removes an `SP0016` if the message attaches to a rule the document actually states — and usually it does not:

| Rule the message attaches to | Count | Has a declarative counterpart |
|---|---|---|
| `Must` / `MustAsync` | ~248 | no |
| `NotEqual`, `InclusiveBetween`, `IsInEnum` | ~37 | no |
| `NotEmpty`, `MaximumLength`, `GreaterThan(OrEqualTo)`, `Matches`, `LessThanOrEqualTo`, `Equal`, `EmailAddress` | ~123 | yes |

For the ~285 whose *rule* lives in code, the rule is dropped and the message has nothing to attach to. Those still report `SP0016`, but now name `$strings.InvoicingMessages.Foo` rather than raw lambda text, so what was lost is inspectable.

133 recovered against ~123 countable by regex — which under-counts multi-line chains — means **every message reachable by this change was recovered**. Closing the rest requires `Must` to gain a declarative counterpart, which is a language question, not more message recovery.

## Verification

- Release build with a file logger across 3 TFMs: 0 errors, 0 warnings.
- **1121 specs** (1106 before, +15). Five confirmed to fail against pre-fix code; the perimeter facts — composed message stays dropped, unwritable key stays dropped, plain constant still recovers — pass either way, correctly, since they guard unchanged behaviour.
- End-to-end gate: exit 0, reads back clean.
- Reference application, measured back-to-back: `SP0016` **643 → 510**, both runs read back clean, and the document goes from **0** to **121** lines carrying `message $strings.…` across 10 resource classes.
- Determinism: byte-identical across consecutive runs.

**Caveat:** that application is under active concurrent edit — a trial run before the measurement pair showed a different baseline entirely (9,164 lines vs 9,209). The reported pair ran back-to-back in one shell invocation, and every non-`SP0016` count is identical between them, which is what isolates the delta to this change rather than to the moving source.

woksin:
Consolidated reviewer context for the whole change. This PR supersedes #2408, #2410 and #2411; all 34 commits are here.

## Where to look first

**The `SP0024` / `SP0034` suppression interaction.** `SP0024` (source did not compile) previously suppressed `SP0034` (the generator reading its own printed document back) unconditionally. It is now narrowed to the Error branch. As a Warning, `SP0024` says the model stands — and suppressing `SP0034` there would hand back a document the Screenplay compiler *rejects* while reporting `IsSuccess == true`, the one outcome `SP0034` exists to prevent. Two changes interacting in a way neither issue anticipated.

**The end-to-end CI gate** (`Source/DotNET/Screenplay.EndToEnd/`) is the most valuable thing here and deliberately is not in the release notes. It generates from `TestApps/AspNetCore` and compiles the result with the `Cratis.Screenplay` compiler, failing on **any** compiler diagnostic including warnings. Note that because the workflow triggers on `pull_request_target`, a job added in a PR runs from the *base* branch and cannot execute on its own PR — it was proven by dispatching the workflow against the branch directly, where it generated 53 lines and reported "reads back clean" on a runner.

**Specification volume.** The document roughly doubles: 5,348 → 9,263 lines on the reference application, of which ~3,900 are the 507 recovered specifications. That is ~7 lines per scenario rather than the explosion feared, because most values are runtime identities and get dropped. **No option to disable was added**, on the reasoning that if a larger application does struggle, the right lever is a per-slice filter driven by a real complaint rather than a global switch added on speculation. Whether 507 scenarios *belong* in a document meant to be read is a judgment left to you.

## Four of the issues rested on wrong evidence

All checked against source; corrections posted to the issues themselves.

- **#2400** — only 7 of the 15 named concepts are recoverable. `FirstName`/`LastName` appear nowhere outside their own declaration files; `PhoneNumber` only in an interface method signature. They are unreferenced, not dropped. `PersonalDetails` carries `FullName`/`DisplayName`/`Bio`. Personal data did improve where reachable: `@pii` 27 → 29, concepts 105 → 140.
- **#2397** — all 510 `WithMessage` calls point at resx designer *properties* (runtime culture lookups), so constant recovery cannot reach them. Filed **#2409** for the `$strings.` route, implemented here.
- **#2403** — 6 → 2, not 6 → 0. Two are a method call and a `static readonly`, correctly still dropped.
- **#2404** — the documented `EventLog.Append`/`ExecuteCommand` shapes do not occur in the reference application at all; it uses `Scenario.Execute` exclusively. Supporting only the documented shape would have recovered nothing. Both are supported.

And one estimate of mine was wrong: **#2409 predicted `SP0016` would fall by ~247; it falls by 133.** Every readable message was recovered — but ~285 attach to `Must`/`MustAsync` and friends, which have no declarative counterpart, so the rule is dropped and the message has nothing to attach to. Closing those is a language question.

## Defects found that no issue described

- A conditional given was read as **unconditional**, stating a world nobody specified. Cost 29 specifications to fix; a document that says less and is true.
- A command reaching an aggregate root in another project crashed with `SyntaxTree is not part of the compilation` — already reachable on the single-project path.
- Two parallel branches claimed `SP0037`/`SP0038` for different meanings. Diagnostic codes are a consumer-facing contract that is never reused, so this was resolved before merge: multi-project keeps them, specifications took `SP0039`/`SP0040`.

## Decisions taken against the issues, deliberately

- **#2398's type-name preference declined** — one `IConstraint` may declare several rules in a single `Define`, so naming them all after the type collapses them into duplicate `constraint` names in a slice. On the reference application it also buys nothing.
- **No new diagnostic codes for #2409.** `SP0041`–`SP0043` were allocated and went unused: the remaining drops are still "a rule that could not be expressed declaratively", which `SP0016` already says.
- **`$strings.` keys are qualified** (`$strings.<Class>.<Key>`) because 6 keys genuinely collide across two resource classes carrying different text. The key is taken from the `GetString` literal, not the property name, because `StronglyTypedResourceBuilder` mangles non-identifier characters.

## Diagnostics on the reference application

| Code | Plan baseline | After | |
|---|---|---|---|
| SP0011 | 6 | **2** | property-reference operands emitted |
| SP0016 | 643 | **510** | resource keys recovered |
| SP0021 | 1 | **0** | imported, and gone entirely under multi-project |
| SP0024 | 1 | **0** | host supplies build-generated compile items |
| SP0025 | 3 | **0** | `obj/` folders excluded from slice attribution |
| SP0031 | 0 | 1 | newly reachable enum name collision, reported |
| SP0035/36 | 0 | 13 / 29 | undeclarable record shapes; cross-slice bindings previously silent |
| SP0039/40 | 0 | 136 / 1554 | unreadable scenarios and scenario values, previously not attempted |

## Verification

- Full `Arc.slnx` clean Release build: 0 errors, 1 pre-existing `CS0436` in `TestApps/Chronicle` that reproduces on `main`.
- **1121 specs**, from 690 when this line of work began. Each phase's additions were confirmed to fail against pre-fix code by reverting and re-running.
- Full Arc suite passes; two `Arc.Core.Specs` failures per full run are the known parallel flake — 1552/1552 isolated.
- Reference application, **whole solution**: reads back clean under the official Screenplay compiler, **0** `import` lines, 229–230 `file` paths all relative and rooted per project, byte-identical across consecutive runs.

**Caveat on every measurement above:** that application is under active concurrent edit — absolute line counts drift by tens of lines between runs. Every before/after pair was taken back-to-back in one shell invocation, which is what isolates a delta to the change rather than to the moving source.

woksin:
One addition to the reviewer notes above, on the weakest content in the generated document.

**347 of the 507 recovered specifications end in `then error ""`** — an empty reason. That is roughly two thirds of them, and many are as thin as:

```
specification WhenExchangingAndMagicLinkIsNotActive
  when ExchangeToken
  then error ""
```

This is **forced by the grammar, not a shortcut**: `then error` requires a `StringLiteral` (`grammar.md:299`, and the parser regex requires the quotes). The source asserts only `IsSuccess.ShouldBeFalse()`, which names no reason, and inventing "rejected" or "validation failed" would put a sentence in the document that the application never states.

Filed upstream as **Cratis/Screenplay#46**, proposing a bare `then error` meaning "rejected, for a reason the specification does not name" — keeping `then error "…"` for the case where a reason genuinely is named. Once that lands, these 347 lines become honest rather than blank-looking, with no change to what is recovered.

Until then the options are to keep them as they are, or to skip scenarios whose only assertion is an unnamed failure — a one-line change that would drop those 347 from the document. Kept as they are, on the grounds that the scenario name carries the meaning and "it fails" is true; happy to switch if you would rather the document did not carry them.

woksin:
Superseded by #2418, which now targets `main` directly and carries all 38 commits — everything here plus the end-to-end gate hardening — as one change with combined release notes.

Nothing is lost: every commit here is an ancestor of that branch, including the documentation correction, and the branch is kept. Reviewer context for all of it is consolidated into a single comment there.

Closing so there is one Arc PR to review rather than two that have to be merged in order.

Issue #2381, opened 2026-07-27
Proxy generator drops roles declared via [Roles("...")] constructor form


woksin:
## Describe the bug

`MethodInfoExtensions.GetRolesFromAttributesData` in the proxy generator only reads roles from an attribute's **named** arguments, but the canonical `[Roles(...)]` attribute sets `Roles` from its **constructor**. As a result `[Roles("Librarian")]` — the documented, idiomatic form — yields no roles at all, silently.

`Source/DotNET/Tools/ProxyGenerator/MethodInfoExtensions.cs:96-117`:

```csharp
var rolesArg = attr.NamedArguments.FirstOrDefault(a => a.MemberName == "Roles");
if (rolesArg != default && rolesArg.TypedValue.Value is string rolesStr && !string.IsNullOrEmpty(rolesStr))
{
    foreach (var role in rolesStr.Split(',', ...))
    {
        yield return role;
    }
}
```

`Source/DotNET/Arc.Core/Authorization/RolesAttribute.cs:16-19`:

```csharp
public RolesAttribute(params string[] roles)
{
    Roles = string.Join(',', roles);
}
```

Because `RolesAttribute` derives from `AuthorizeAttribute`, it *is* correctly recognised as an authorization attribute — the `isAuthorizeAttribute` check passes. It then finds no `Roles` named argument and yields nothing. So the failure is silent: authorization metadata is simply absent from the generated proxies rather than erroring.

`[Authorize(Roles = "Librarian")]` (named-argument form) works, which is likely why this has gone unnoticed.

## Steps to reproduce

1. Annotate a command or query with the idiomatic attribute:
   ```csharp
   [Roles("Librarian")]
   [Command]
   public record RegisterAuthor(AuthorName Name) { ... }
   ```
2. Run proxy generation.
3. Inspect the generated TypeScript proxy's authorization metadata.

## Expected behavior

`[Roles("Librarian")]` and `[Roles("Librarian", "Admin")]` produce the roles `Librarian` and `Librarian, Admin` respectively — the same result as `[Authorize(Roles = "Librarian")]`.

## Actual behavior

No roles are emitted. The command or query appears to have no role requirement.

## Environment

- OS: macOS 15
- .NET version: 10.0.x
- Arc version: 18.7.x (current `main`)

## Additional context

[...]

PR #2417, opened 2026-07-29
Screenplay generation and proxy generator hardening fixes

woksin:
# Summary

A batch of small, well-scoped fixes from an open-issue triage sweep. Each is an independent commit.

> **Reviewer note:** these changes were authored in an environment without a .NET SDK (the SDK download host was blocked), so they could not be compiled or tested locally. They were verified by careful inspection against the actual codebase APIs and follow existing spec conventions, but rely on CI for build/test confirmation. Please let CI validate before merge.

## Fixed

- Proxy generator now reads roles declared via the `[Roles("...")]` constructor form, not only the named-argument form (#2381)
- Screenplay generation no longer treats generated files under `obj/`/`bin/` as slice source folders, removing false `SP0025` diagnostics (#2395)
- Screenplay generation renders separator-carrying (kebab/snake-case) constraint names as readable PascalCase identifiers instead of run-together text (#2398)
- Screenplay generation recovers validation messages from `WithMessage(_ => Constant)` lambda forms instead of dropping them as `SP0016` (#2397)

---

### DIFF
--- a/Source/DotNET/Screenplay.Specs/for_ApplicationModelAnalyzer/when_analyzing/a_slice_whose_generated_members_sit_under_obj.cs
+++ b/Source/DotNET/Screenplay.Specs/for_ApplicationModelAnalyzer/when_analyzing/a_slice_whose_generated_members_sit_under_obj.cs
@@ -1,0 +1,59 @@
+// Copyright (c) Cratis. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Cratis.Arc.Screenplay.Analysis;
+using Cratis.Arc.Screenplay.Model;
+
+namespace Cratis.Arc.Screenplay.for_ApplicationModelAnalyzer.when_analyzing;
+
+/// <summary>
+/// A source generator writes partial members into a slice's namespace and emits them to disk under <c>obj/</c>. Those
+/// files contribute symbols to the slice but say nothing about where its source - and therefore its screens - live, so
+/// the slice is not spread over the folder they sit in and screen discovery must not scan it.
+/// </summary>
+public class a_slice_whose_generated_members_sit_under_obj : Specification
+{
+    const string Source = """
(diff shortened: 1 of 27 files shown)


woksin:
Closing this as superseded, with its unique commit preserved rather than discarded.

**Four of the five commits here are already implemented in #2418**, in the same files, so the two cannot both merge cleanly:

| | here | #2418 |
|---|---|---|
| #2395 `obj/`/`bin/` in slice attribution | ✅ | ✅ with specs and a measured `SP0025` 3 → 0 |
| #2398 separator-carrying names | ✅ | ✅ with specs, 31 names verified against a real application |
| #2397 `WithMessage` lambda constants | ✅ | ✅ with specs |
| #2400 concepts via child records | ✅ | ✅ with specs, concepts 105 → 140, `@pii` 27 → 29 |
| #2381 proxy generator `[Roles("…")]` | ✅ | ❌ |

**#2381 has been extracted to #2419**, rebased onto `main`. It did not compile as written — `CS0104`, because `Cratis.Arc.Authorization` and `Microsoft.AspNetCore.Authorization` both declare an `Authorize` attribute and both were imported; then `CS1614` when aliased; then `RCS1181` in Release. All three are fixed there, and the four `when_extracting_roles` specs pass on net8.0 and net9.0. The production change itself was correct as written.

**One correction worth carrying over.** The `#2397` bullet here says messages are recovered "instead of dropping them as `SP0016`". On a real production application that recovers **nothing**: all 510 `WithMessage` calls resolve to resx-generated designer *properties*, which are runtime culture-dependent lookups with no constant value, so `HasConstantValue` cannot reach them. That finding is why **#2409** was filed and implemented in #2418, emitting `$strings.` keys instead. Measured there: `SP0016` 643 → 510.

Also, the description carries an AI-attribution footer, which the repository's PR guidance asks not to include.

Nothing is lost by closing this — the branch is kept, four commits are superseded by tested equivalents, and the fifth is open as #2419.

Issue #2416, opened 2026-07-29
verify-markdown.sh reports success while scanning zero links


woksin:
`Documentation/verify-markdown.sh` reports success, but its link-checking step is checking nothing.

Observed while adding a page: the linkinator step prints

```
Successfully scanned 0 links
```

both before and after adding a page containing nine links, two of them new anchors. The lint step (259 files, 0 issues) works; the link step does not.

### Why it matters

The documentation rules state that every link must resolve and that broken links fail review, and `verify-markdown.sh` is the mechanism that is supposed to enforce it. A check that reports success while scanning zero links is worse than no check — it produces a green result that a contributor and a reviewer both reasonably read as "the links are fine". Links were verified by hand for the page in question, which does not scale.

This is the same failure mode as a test suite that silently matches nothing: the signal and the absence of signal look identical.

### Suggested direction

Find why linkinator resolves no files — most likely a glob or working-directory mismatch between the script and where the markdown actually lives — and fix it so the count is non-zero. Then add a guard: **fail the script if the scanned-link count is zero**, so the check cannot silently become a no-op again. A deliberately broken link in a scratch file is the cheapest way to prove the check works before trusting it.

### Acceptance

`verify-markdown.sh` reports a non-zero scanned-link count on this repository; a deliberately broken internal link makes it fail; and the script fails rather than passes if it ever scans zero links again.

woksin:
## Resolved on `main`

The link step no longer scans zero links, and can no longer silently become a no-op again.

`Documentation/verify-links.sh` now:

- reads the scanned-link count out of linkinator's summary (stripping colour codes first, and matching mid-word so both "Successfully scanned N links" and "Scanned N links" are read);
- **fails when the count is zero**, and separately when no summary was produced at all ("the file globs most likely matched nothing");
- pins `linkinator@8.0.2` so a new major cannot change what is scanned;
- uses one `Documentation/**/*.{md,mdx}` glob, because linkinator aborts on any glob matching nothing;
- skips site-absolute links by rule rather than by name, with the comment recording the root cause: *"The pattern must never match the crawl root itself — a pattern that did is what turned this check into a no-op."*

That last line is the actual bug this issue reported. Both acceptance criteria are met.

@einari @woksin — safe to close.

woksin:
Closing — released in **v20.66.0**.

Delivered by commit `a9825914` ("Check the documentation links instead of reporting success on nothing"). Verification against `main` is in the comment above.

Issue #2415, opened 2026-07-29
Screenplay generation: query paging, sorting, routes and read model tags are dropped silently


woksin:
The generator's stated contract is that nothing is dropped silently — everything it cannot express emits a located `SP00xx` diagnostic, so a reader can tell the difference between "the application does not do this" and "the generator could not say it".

Three things break that contract today. They are dropped with **no diagnostic at all**:

- **Query paging and sorting.** `a_query_the_host_hands_more_than_arguments.cs` asserts `should_report_nothing`.
- **Custom route templates.** A command or query with an explicit route has it dropped without a word.
- **Read model tags.** Note *event* tags are recovered and printed, so a reader seeing tags on events could reasonably conclude a read model simply has none.

This was found while correcting `Documentation/generating-a-screenplay.md`, which asserted "Each is reported as a diagnostic when encountered" for the whole table — an assertion that was not true for these three rows. The documentation is now accurate, but the accurate statement is that the generator has silent drops, which is the thing the contract exists to prevent.

### Why it matters

A silent drop is worse than a loud one in a document whose only value is describing the real application. A reader has no way to know the question was even asked. Two of the defects found before this package shipped were exactly this shape — an event property named `Tag` silently absorbed, and `authorize`/`produces` collisions silently swallowing a property — and both were fixed by making them report.

### Suggested direction

Report each at Information severity where the construct is recognised but has no counterpart:

- paging and sorting: the document describes the model, not how it is served, so this is a genuine non-counterpart rather than a gap in reading
- custom routes: same
- read model tags: worth distinguishing, since a read model has no declaration of its own in the document — it appears only as the type a query returns — so this may be better framed as "the read model itself is not declared" than as a dropped tag

Whether these deserve one shared code or three is a judgment call. A single code covering "a transport or storage concern the document does not describe" may be more useful to suppress on than three.

### Acceptance

Each of the three emits a located diagnostic; specs cover all three; the counts on a real application are reported so the volume is known before it ships (read model tags in particular could be numerous).

woksin:
## Resolved on `main`

All three now report. `Source/DotNET/Screenplay/ScreenplayDiagnosticCodes.cs` (commit `69823198`, "Report what an application says about how it is served") added:

- **`SP0041` `ServingConcernWithoutCounterpart`** — paging, sorting and custom route templates, sharing one code because they share one reason, which is the judgment call the issue asked for.
[...]

PR #2418, opened 2026-07-29
Recover more of an application, describe it across projects, and report what cannot be recovered

woksin:
## Added

- A validation message looked up from a resource is written as the key it is looked up by, so the document states which message a rule carries without baking one culture into it (#2409)

## Changed

- A specification value that is a freshly generated identity is no longer reported as unreadable, since an identity the source generates states nothing for the document to carry (#2404)

## Fixed


woksin:

**This is agent-authored code that no human has read yet.** Green CI is not sign-off. **Do not merge without explicit review.** Stacked on #2413; deliberately kept as its own PR rather than folded into it, because #2413 is already 35 commits and 146 files.

## Why this exists

The end-to-end gate generated from a **53-line** document. Every defect that mattered in this line of work was found on a real 9,000-line application, not there: the `Description` collision, the silently absorbed `Tag` property, enum constants printed as integers, path roots anchored on the wrong directory, a conditional `given` read as unconditional. A gate that cannot fail on any of those is closer to decoration than protection.

`TestApps` now carries those shapes, and CI reads back **both** applications rather than one.

## The gate was proven able to fail, not assumed to be

Three deliberate breakages, each a one-line removal from `ReservedWords`, rebuilt and restored:

| Shape | Broken | Restored |
|---|---|---|
| `description` | exit **1** — `Invalid description 'description String'` | exit **0** |
| `concurrency` | exit **1** — `Invalid concurrency declaration 'concurrency Int'` | exit **0** |
| `tag` | exit **1** — `Invalid tag value 'ApprovalLabel[]'` | exit **0** |

The `Tag` fixture is a **collection** on purpose: a swallowed `tag X[]` is rejected by the language, whereas `tag X` compiles silently — so the collection form is the one that can move an exit code.

## What the gate still cannot catch, stated plainly

**The enum-as-integer and undeclared-concept regressions do not move the exit code.** The shipped Screenplay compiler was probed directly: `set status to 6` where the concept declares `pending`/`granted`/`declined` produces **no diagnostic**, and a reference to a concept the document never declares produces **no diagnostic**. Both shapes are in the fixture, so a regression shows up as a changed document and changed diagnostics that a reviewer can see — but CI stays green. Closing that needs a committed golden document, which is a larger change and was not made.

## A correction to the premise I was given

The diagnostic firing 1554 times is **`SP0040`** (`UnreadableSpecificationValue`), not `SP0038` (`ProjectsWithoutASharedRoot`). The description identified it; the number did not.

**Decision: recognise a freshly generated identity and do not report it** — rather than aggregate. Per-value reporting is what lets a reader count what a scenario states against what the source states; aggregating would fold a non-difference in with the real ones instead of removing it. A fresh identity has nothing to state, so leaving it out says exactly as much as the source does.

Recognition is narrow: a static, parameterless method named `New`/`NewGuid` returning the type that declares it, where that type is `System.Guid` or derives from `ConceptAs<T>` — a method with no input has nothing to derive a value from. It follows an identifier back through the field initializer a scenario holds an identity in, since two steps agreeing on one identity is why it is held rather than written twice.

**On the reference application: `SP0040` 1554 → 912, document byte-identical.** The remaining 912 are different in kind and should stay — `new DateOnly(2026, 9, 1)`, `["5+ years C#"]`, `string.Empty`, `OrganizationNumber.NotSet`, and derived identities such as `TimesheetId.For(engagementId, year, month)`, which is a value the source really does state.

## Project filtering

The harness filtered candidate projects by name, so a `netstandard2.0` analyzer project loaded as an application project. It now asks whether a compilation can resolve `CommandAttribute` or `EventTypeAttribute` — a project that resolves neither cannot declare anything a document is made of. On `Arc.slnx` that turns away 10 projects and takes 4 of the 6 `SP0024` reports with them; document byte-identical.

**Misses:** a host project that sees the framework but declares nothing is still read and contributes an empty model. The specification-project name filter was **kept** — nothing about what a specs project can *see* distinguishes it, since referencing the framework is the point of it, and replacing that check would pull fixture commands and events into the document as real artifacts.

## Verification

- Release build 0 errors / 0 warnings; full `Arc.slnx` clean Release 0 errors, 1 pre-existing `CS0436`.
- **1132 specs** (1121 before, +11). The two fix-proving assertions were confirmed to fail against pre-fix code; the third is a perimeter guard against over-broad recognition and passes either way by construction.
- Gate on `TestApps/AspNetCore`: exit 0, 64 lines (was 53), `SP0032 x2` — the two new collisions.
- Gate on `TestApps/Chronicle`: exit 0, 75 lines (was 21), including two recovered `specification` blocks.
- Determinism: two consecutive runs on each, byte-identical.

## Disclosure

Three commits were amended once before pushing: the full Release solution build turned up `NU1201` from the new project reference in commit 1, so commits 2 and 3 were saved as patches, reset to commit 1, the `TargetFrameworks` fix amended in, and the patches re-applied — so each commit builds on its own. Nothing had been published at the time; no rebase onto another branch, no force-push.

woksin:
Consolidated reviewer context for the whole change. This PR supersedes #2408, #2410, #2411 and #2413; all 38 commits are here, across 160 files.

**This is agent-authored code that no human has read yet.** Green CI is not sign-off. **Do not merge without explicit review.** It is large — the reviewing order below is meant to make that tractable.

## Read these first

**1. The `SP0024` / `SP0034` suppression interaction.** `SP0024` (source did not compile) previously suppressed `SP0034` (the generator reading its own printed document back) unconditionally. It is now narrowed to the Error branch. As a Warning, `SP0024` says the model stands — and suppressing `SP0034` there would hand back a document the Screenplay compiler *rejects* while reporting `IsSuccess == true`, the one outcome `SP0034` exists to prevent. Two changes interacting in a way neither issue anticipated.

**2. The end-to-end gate, and what it still cannot catch.** `Source/DotNET/Screenplay.EndToEnd/` generates from `TestApps/AspNetCore` and `TestApps/Chronicle` and compiles both, failing on any compiler diagnostic including warnings. It was proven able to fail rather than assumed to be — three deliberate breakages, each exiting 1 and returning to 0 on restore:

| Shape | Broken | Restored |
|---|---|---|
| `description` | exit 1 — `Invalid description 'description String'` | exit 0 |
| `concurrency` | exit 1 — `Invalid concurrency declaration 'concurrency Int'` | exit 0 |
| `tag` | exit 1 — `Invalid tag value 'ApprovalLabel[]'` | exit 0 |

**But the enum-as-integer and undeclared-concept regressions do not move the exit code.** The shipped Screenplay compiler was probed directly: neither produces a diagnostic. Those shapes are in the fixture, so a regression shows as a changed document a reviewer can see, but CI stays green. Closing that needs a committed golden document, deliberately not bolted on here.

Note also that because the workflow triggers on `pull_request_target`, a job added in a PR runs from the *base* branch and cannot execute on its own PR. The gate was proven on a runner by dispatching the workflow against the branch directly.

**3. Specification volume.** The document roughly doubles — 5,348 → 9,263 lines on the reference application, ~3,900 of it the 507 recovered specifications, about seven lines each. **No option to disable was added**, on the reasoning that if a larger application struggles, the right lever is a per-slice filter driven by a real complaint rather than a global switch on speculation. Whether 507 scenarios *belong* in a document meant to be read is a judgment left to you.

Related: **347 of those 507 end in `then error ""`**. That is forced by the grammar — `then error` requires a `StringLiteral` — and the source asserts only that something was rejected, without naming a reason. Inventing one would state something the application never says. Filed upstream as Cratis/Screenplay#46.

## Four of the issues rested on wrong evidence

All checked against source; corrections posted to the issues themselves.

- **#2400** — only 7 of 15 named concepts are recoverable. `FirstName`/`LastName` appear nowhere outside their own declaration files; `PhoneNumber` only in an interface method signature. They are unreferenced, not dropped. Personal data still improved where reachable: `@pii` 27 → 29, concepts 105 → 140.
- **#2397** — all 510 `WithMessage` calls point at resx designer *properties*, which are runtime culture lookups with no constant value. Filed **#2409** for the `$strings.` route, implemented here.
- **#2403** — 6 → 2, not 6 → 0.
- **#2404** — the documented `EventLog.Append`/`ExecuteCommand` shapes do not occur in the reference application at all; it uses `Scenario.Execute` exclusively. Supporting only the documented shape would have recovered nothing.

And one estimate of mine was wrong: **#2409 predicted `SP0016` would fall by ~247; it falls by 133.** Every readable message was recovered, but ~285 attach to `Must`/`MustAsync`, which have no declarative counterpart — filed as Cratis/Screenplay#47.

## Defects found that no issue described

- A conditional `given` was read as **unconditional**, stating a world nobody specified. Cost 29 specifications to fix.
- A command reaching an aggregate root in another project crashed — already reachable on the single-project path.
- Two parallel branches claimed `SP0037`/`SP0038` for different meanings; resolved before merge, since diagnostic codes are a contract that is never reused.
- The documentation asserted model-bound `[ChildrenFrom]`/`[Nested]` were "not read back yet" — false since #2385 — and claimed every unexpressible thing is reported, which is **not** true for query paging, sorting, routes and read model tags. Documentation corrected here; the silent drops are filed as #2415.

## Deliberate decisions against the issues

- **#2398's type-name preference declined** — one `IConstraint` may declare several rules in a single `Define`, so naming them all after the type collapses them into duplicate names.
- **No new diagnostic codes for #2409.** `SP0041`–`SP0043` were allocated and went unused; the remaining drops are still what `SP0016` already says.
- **`$strings.` keys are qualified**, because 6 keys collide across two resource classes with different text. The key comes from the `GetString` literal, not the property name, which `StronglyTypedResourceBuilder` mangles.
- **`SP0040` recognises a freshly generated identity** rather than aggregating: 1554 → 912 on the reference application, document byte-identical. The remaining 912 are different in kind — `new DateOnly(2026, 9, 1)`, `TimesheetId.For(...)` — and should stay.

## Diagnostics on the reference application

| Code | Plan baseline | After | |
|---|---|---|---|
| SP0011 | 6 | **2** | property-reference operands emitted |
| SP0016 | 643 | **510** | resource keys recovered |
| SP0021 | 1 | **0** | imported, and gone entirely under multi-project |
| SP0024 | 1 | **0** | host supplies build-generated compile items |
| SP0025 | 3 | **0** | `obj/` folders excluded from slice attribution |
| SP0035/36 | 0 | 13 / 29 | undeclarable record shapes; cross-slice bindings previously silent |
| SP0039/40 | 0 | 136 / 912 | unreadable scenarios and values, previously not attempted |

## Verification

- Full `Arc.slnx` clean Release: 0 errors, 1 pre-existing `CS0436` in `TestApps/Chronicle` that reproduces on `main`.
- **1132 specs**, from 690 when this work began. Each phase's additions were confirmed to fail against pre-fix code by reverting and re-running.
- Full Arc suite passes; two `Arc.Core.Specs` failures per full run are the known parallel flake — 1552/1552 isolated.
- Reference application, whole solution: reads back clean under the official Screenplay compiler, **0** `import` lines, all `file` paths relative and rooted per project, byte-identical across consecutive runs.
- `Documentation/verify-markdown.sh` passes — though its link step scans zero links, filed as #2416.

**Caveat on every measurement:** that application is under active concurrent edit; absolute counts drift between runs. Every before/after pair was taken back-to-back in one shell invocation, which is what isolates a delta to the change rather than to the moving source.

PR #2419, opened 2026-07-29
Read roles declared with the Roles constructor form

woksin:
## Fixed

- Roles declared with the `[Roles("...")]` constructor form are read by the proxy generator, not only the named-argument form (#2381)

### DIFF
--- a/Source/DotNET/Tools/ProxyGenerator.Specs/ControllerBased/for_MethodInfoExtensions/RoleSecuredTypes.cs
+++ b/Source/DotNET/Tools/ProxyGenerator.Specs/ControllerBased/for_MethodInfoExtensions/RoleSecuredTypes.cs
@@ -1,0 +1,23 @@
+// Copyright (c) Cratis. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Cratis.Arc.Authorization;
+using Microsoft.AspNetCore.Authorization;
+
+namespace Cratis.Arc.ProxyGenerator.ControllerBased.for_MethodInfoExtensions;
+
+public class RoleSecuredTypes
+{
+    [Roles("Librarian")]
+    public void SingleRoleFromConstructor() { }
+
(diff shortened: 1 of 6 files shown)


woksin:

**Provenance.** This is one commit extracted from #2417 (`43ac8e46`), rebased onto `main` and made to compile. It is the only part of that PR not already implemented in #2418 — the Screenplay fixes there (#2395, #2397, #2398, #2400) are all covered by #2418, with specs and measurements against a real application.

**It did not compile as written.** #2417's reviewer note says it was authored in an environment without a .NET SDK and relies on CI. Cherry-picked onto `main`, the fixture failed with `CS0104: 'Authorize' is an ambiguous reference` — `Cratis.Arc.Authorization` and `Microsoft.AspNetCore.Authorization` both declare an `Authorize` attribute and both were imported. Aliasing it then failed with `CS1614` (ambiguous between the attribute and its `Attribute`-suffixed form), so the ASP.NET Core attribute is now written out in full where the fixture needs its named-argument form. A third failure followed in Release only — `RCS1181`, comment must be a documentation comment — which is the usual reminder that Release is the real gate here, since analyzers are errors only there.

The production change itself (`MethodInfoExtensions.cs`, reading `ConstructorArguments` as well as `NamedArguments`) was correct as written and needed no change.

**This is agent-authored code that no human has read.** **Do not merge without explicit review.**

**Verification**

- `ProxyGenerator.csproj` Release: 0 errors, 0 warnings.
- `ProxyGenerator.Specs` Release: the four new `when_extracting_roles` specs pass on net8.0 and net9.0 — constructor form with one role, with several roles, the named-argument form, and both forms combined.

**Note on #2417.** Four of its five commits duplicate work already in #2418, in overlapping files, so the two cannot both merge cleanly. Its `#2397` release note also overstates the effect: on a real application that recovers nothing, because every `WithMessage` there resolves to a resx designer property with no constant value — which is why #2409 exists and is implemented in #2418.

PR #2420, opened 2026-07-29
Align tenancy docs with the actual default tenant header

woksin:
## Fixed

- Tenancy documentation no longer presents `X-Tenant-ID` as the tenant header; the header-override samples now use a clearly-custom placeholder so they no longer contradict the actual default `x-cratis-tenant-id` (#2351)

---


woksin:

PR #2421, opened 2026-07-29
Check the documentation links instead of reporting success on nothing

woksin:
## Fixed

- Documentation link verification checks links again — it reported success while scanning zero links, both locally and in CI (#2416)
- Fourteen documentation links that named pages under a file name those pages no longer have

### DIFF
--- a/.github/workflows/markdown-verification.yml
+++ b/.github/workflows/markdown-verification.yml
@@ -33,12 +33,12 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
 
-      - name: Check links in Documentation
-        uses: JustinBeckwith/linkinator-action@v1
+      - name: Setup Node
+        uses: actions/setup-node@v4
         with:
-          paths: 'Documentation/**/*.md'
-          markdown: true
-          recurse: true
-          redirects: allow
-          statusCodes: '403:ok'
-          linksToSkip: '^(https?:\/\/)?(localhost|127\.0\.0\.1)(:\d+)?(\/|$)'
+          node-version: 23.x
+
+      # The same script contributors run locally, so that what CI checks and
+      # what verify-markdown.sh checks cannot drift apart.
+      - name: Check links in Documentation
+        run: ./Documentation/verify-links.sh
--- a/Documentation/backend/chronicle/index.md
+++ b/Documentation/backend/chronicle/index.md
@@ -45,7 +45,7 @@ Without this package, Arc and Chronicle are independent. With it:
 | Topic | Description |
 | ----- | ----------- |
 | [Aggregates](aggregates/index.md) | Working with aggregate roots and event sourcing. |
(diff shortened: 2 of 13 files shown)


woksin:

**This is agent-authored and no human has read it yet.** Green CI is not sign-off.

## The cause was not the glob

The issue guessed a glob or working-directory mismatch. It is neither. `linkinator` serves the files it scans from a local web server on `127.0.0.1`, so every internal link resolves to `http://127.0.0.1:<port>/…`. The skip pattern `^(https?:\/\/)?(localhost|127\.0\.0\.1)(:\d+)?(\/|$)` therefore matched **the crawl root itself**, the whole scan was skipped, and the run reported success:

```
✓ Successfully scanned 0 links in 0.011 seconds.   (exit 0)
```

The same pattern is passed to `linkinator-action` in the workflow, so CI was equally blind — run 30450036709 logs `Scanned total of 0 links!`.

## What is verified here and what is not

A single repository can only resolve relative links and external URLs. Links written as site-absolute paths (`/arc/…`, `/chronicle/…`) address the aggregated site, where each product's `Documentation` folder is mounted under its own prefix; they are verified when that site is built. The skip is now expressed as a rule — "served by our own scan server and outside `/Documentation/`" — so a new product prefix needs no change here.

## Verification

| | Before | After |
|---|---|---|
| links scanned | 0 | 317 |
| broken links reported | 0 | 0 (14 found and fixed) |

- **Pre-fix configuration passes with a broken link present** — a scratch page with a dead link, old flags: `Successfully scanned 0 links`, exit **0**.
- **New script fails on the same page** — `Detected 1 broken links. Scanned 319 links`, exit **1**.
- **Zero-scan guard** — restoring the old skip pattern in a copy of the script now exits **1** with `Link verification scanned zero links`, rather than passing.
- **No-summary guard** — globs matching nothing exits **1** rather than passing.
- `Documentation/verify-markdown.sh` passes end to end (lint 259 files + links 317), from the repository root and from `Documentation/`.

## The fourteen links

Five kept `.md` after the page became `.mdx` (`add-event-sourcing`, `your-first-command`, `frontend/index`, `storybook`) — those resolve nowhere, on the site or off it. Nine were extensionless chapter links in the tutorial, which the site resolves but a file-level check cannot; they now carry the file name like the rest of the documentation does (`generating-a-screenplay.md` already links `.mdx` this way).

## Worth knowing

- Every folder under `Documentation/` currently has an `index` page, so `--directory-listing` is equivalent to a strict check today. If a folder ever loses its index page, a link to that folder would pass here — the aggregated site build is what catches that.
- `linkinator` is pinned to 8.0.2 so a new major cannot silently change what is scanned.
- `.mdx` pages are now scanned for links (13 files, +18 links). They are still not linted — `markdownlint` runs on `*.md` only, unchanged by this PR.

woksin:

PR #2423, opened 2026-07-29
Inject read models from non-Chronicle providers into the command pipeline

woksin:
# Summary

Implements the design proposed in #2303 so an EF Core-backed `[ReadModel]` can be injected by key into a command, not just Chronicle-backed ones.

> **Reviewer notes** — this was authored without a local .NET SDK, so it's verified by inspection + CI only, and it touches Chronicle's command-context plumbing, so a couple of decisions are worth your eyes:
> - **Provider-neutral key (the crux from #2303):** the resolved key is lifted into Arc.Core as `Cratis.Arc.Commands.CommandContextKeys.ResolvedKey` (a plain string); Chronicle's `EventSourceValuesProvider` now writes it alongside its existing `EventSourceId`, and EF reads only this Arc-level string. Chronicle behavior is otherwise unchanged.
> - **Two deviations from the prose proposal:** (1) resolvers are handed to `AddReadModelsForCommand` at registration rather than discovered via `IInstancesOf<>` — registering a scoped by-key factory per read model type needs the type list at registration, which each provider already computes; the shared abstraction + additive union are preserved. (2) The key holder is `CommandContextKeys` (not `WellKnownCommandContextKeys`) to avoid a `CS0104` clash with Chronicle's identically-named type.
> - **Known boundary:** only Chronicle currently writes the neutral key, so a pure EF-without-Chronicle app wouldn't have it populated. The #2303 coexistence scenario always includes Chronicle; a dedicated writer can be added if pure-EF is a target.
> - Points I'd most want CI/your review to confirm: `DbContext.Find` on a read-only/no-tracking context, the `ConceptAs<T>` primary-key conversion path, and the SQLite in-memory spec lifetime.

## Added

- Read models backed by non-Chronicle providers (such as Entity Framework Core) can be injected by key into a command's `Handle()`, `Provide()`, and `CommandValidator<>`, through a provider-agnostic resolver abstraction; Chronicle and EF-backed read models coexist without either claiming the other's types, and nullable/non-nullable resolution behaves exactly as before (#2303)

---


woksin:

PR #2425, opened 2026-07-29
Inject read models from any provider, and ship the open work as one release

woksin:
# Summary

The open pull requests are merged here so they ship as one release rather than several, and `main` builds again.

## Added

- A read model backed by Entity Framework Core or MongoDB can be injected by key into a command's `Handle()`, `Provide()`, and `CommandValidator<>`, with nothing to wire up — an entity carried by a `ReadOnlyDbContext` resolves by its primary key, a `[ReadModel]` held in MongoDB by its document `_id`. Nullable and non-nullable resolution behave as for a Chronicle-backed read model (#2303)
- `ICanResolveReadModelForCommand`, for making the read models of another store injectable into a command the same way. A provider states whether it owns the read models it reports or can merely load them, so which one resolves a read model does not depend on registration order (#2303)
- `SP0041` reports what an application says about how it is served and a Screenplay has no counterpart for: a query the host pages and sorts, and every route template declared with `[Path]`, `[Route]` or an HTTP verb, including inherited ones (#2415)
- `SP0042` reports the tags a read model carries, which the document has nowhere to hold while an event's tags are printed (#2415)
- Chronicle 16.9.2: `ComplianceMetadataActionFailed` replaces the bare `InvalidOperationException` when applying or releasing compliance metadata fails. On a failed release it names the property, the subject the value was released under, and explains that the value was encrypted under a different subject, pointing at CHR0038 — rather than surfacing only the underlying OpenSSL `oaep decoding error`.
- Chronicle 16.10.0: `[Replay]` on a reactor handler method, marking it as the one to run while the observer is replaying. Only the marked handler runs during a replay; an event type without one behaves exactly as before. (cratis/chronicle#845)
- Chronicle 16.10.0: `IEventStore.Identities` with `Rename(subject, name)`, exposing the identity renaming the kernel already supported to the .NET client. (cratis/chronicle#1684)
- Chronicle 16.10.0: `[JsonSchemaType(typeof(T))]` on a type, overriding how it is represented in the JSON schema generated for events and read models. Compliance metadata and nullability are preserved. (cratis/chronicle#1470)
- Chronicle 16.10.0: `AddCratisChronicleMongoDB()` for Aspire, provisioning a MongoDB container that initiates itself as a single-node replica set and hands back a `directConnection` connection string. (cratis/chronicle#3400)
- Chronicle 16.10.0: `IReactorMiddleware.BeforeInvoke` and `AfterInvoke` overloads that receive the `ReactorId` of the reactor observing the event. Chronicle calls these; they default to the existing overloads, so a middleware written against the previous interface keeps working unchanged. (cratis/chronicle#1009)
- Chronicle 16.10.0: `CompliancePropertyNotFoundInSchema`, thrown when a property being applied or released has no counterpart in the schema. The message names the property path, the compliance subject, and the properties the schema does declare. (cratis/chronicle#690)
- Chronicle 16.10.0: An overview on the `cratis/chronicle` Docker Hub repository covering the published tag variants, ports, quick start and the environment variables that configure the server. (cratis/chronicle#1277)

## Changed

- `Cratis.Chronicle`, `Cratis.Chronicle.AspNetCore` and `Cratis.Chronicle.Testing` are updated from `16.9.1` to `16.10.0`, picking up [16.9.2](https://github.com/Cratis/Chronicle/releases/tag/v16.9.2), [16.9.3](https://github.com/Cratis/Chronicle/releases/tag/v16.9.3), [16.9.4](https://github.com/Cratis/Chronicle/releases/tag/v16.9.4) and [16.10.0](https://github.com/Cratis/Chronicle/releases/tag/v16.10.0)
- Chronicle 16.9.2: `CHR0038` now also reports the fluent `builder.Join<TEvent>(_ => _.On(...))` form, not just the model-bound `[Join<TEvent>]` attribute. A `[PII]` value counts as reaching the read model through an explicit `.Set(...).To(...)` mapping or through AutoMap matching an identically named property.
- Chronicle 16.10.0: Registering event types on connect issues one read and one bulk write for the whole batch instead of several round trips per event type, so startup no longer scales its database round trips with the number of event types. (cratis/chronicle#984)

## Fixed

- `Cratis.Arc.Chronicle.Testing` compiles against Chronicle `16.10.0` again — the in-memory event store behind a command scenario did not implement the two `IEventStore` members that release adds, which is why `20.65.4` has no NuGet packages
- Documentation link verification checks links again — it reported success while scanning zero links, both locally and in CI (#2416)
- Fourteen documentation links that named pages under a file name those pages no longer have
- Tenancy documentation no longer presents `X-Tenant-ID` as the tenant header; the header-override samples now use a clearly-custom placeholder so they no longer contradict the actual default `x-cratis-tenant-id` (#2351)
- Chronicle 16.9.3: `CHR0038` no longer reports a fluent join when AutoMap cannot carry the value across — `[NoAutoMap]` on the read model, and `[NoAutoMap]` on the target property of a positional record, were both still reported. Since the rule is an error, each was a build break on correct code.
- Chronicle 16.9.4: Two links in the concepts docs pointed at `/arc/backend/chronicle/aggregates/aggregate-root.md`, which no longer exists — the Arc page is `defining-an-aggregate-root.md`. This was breaking the Documentation site's internal-links check.
- Chronicle 16.10.0: Reading back a job whose request was not yet assigned no longer fails with a `NullReferenceException`, losing the status and progress the job state was written to record. (cratis/chronicle#3541)
- Chronicle 16.10.0: Reducer failures are written to the client log. A reducer method that throws was reported to the kernel as a failed observation without ever appearing in the client's own log. (cratis/chronicle#1395)
- Chronicle 16.10.0: Applying or releasing compliance metadata for a property the schema does not declare no longer fails with a bare `Sequence contains no matching element`. (cratis/chronicle#690)
- Chronicle 16.10.0: Applying or releasing compliance metadata no longer fails when a schema flattened across inheritance declares the same property name more than once. (cratis/chronicle#690)


woksin:
## What this supersedes

This branch is `origin/main` with #2420, #2421, #2422, #2423 and #2424 merged in (all five merged clean, no conflicts), plus the work below. Those five should be closed rather than merged if this one lands.

## Why the Chronicle update needed a fix, and why nobody saw it

#2424 on its own does not build. Chronicle 16.10.0 adds `IEventStore.Identities` and `IEventStore.ExternalServices`, and `Chronicle.Testing`'s in-memory `EventStoreForScenario` implements `IEventStore` — two `CS0535`s the moment the package moves.

It went unnoticed because `.NET Build` only triggered on `Source/DotNET/**`, `Specifications/**` and its own workflow file. #2424 changes `Directory.Packages.props` alone, so the only workflow that ever ran on it was CodeQL. The workflow now also watches `Directory.Build.props`, `Directory.Packages*.props`, `global.json`, `Arc.slnx` and `TestApps/**` — what the build compiles *against*, not only what it compiles. Worth knowing that every past dependency-bump PR merged unbuilt for the same reason.

## Design changes to #2423

Two providers can be able to load the same read model, and `AddReadModelsForCommand` decided the winner by registration order (`RemoveAll` then `AddScoped`, last one wins). With MongoDB added that stops being cosmetic: MongoDB serves a collection for any `[ReadModel]`, including every read model Chronicle projects — and Chronicle is the only provider that releases compliance-protected values, so a Mongo-resolved Chronicle read model would hand `[PII]` back still encrypted.

A resolver now declares a `ReadModelForCommandOwnership`. Chronicle and EF Core are `Declared` (a projection/reducer targets it; a `DbSet` carries it) and claim their types outright. MongoDB is `Fallback` and claims only types nothing else resolves — including skipping a type the application registered itself, which the previous unconditional `RemoveAll` would have silently dropped. A declaring provider therefore wins in either registration order; specs cover both orderings.

`Documentation/backend/chronicle/read-models/injecting-into-commands.md` gains the MongoDB section and a table of which provider owns what.

## Known boundary, unchanged from #2423

Only Chronicle writes the provider-neutral `CommandContextKeys.ResolvedKey`, so an application with EF Core or MongoDB and *no* Chronicle has no key to resolve by, and injection fails as a validation error (HTTP 400) rather than resolving. Every key-resolution rule Arc has — `ICanProvideEventSourceId`, `[Key]`, `EventSourceId`-convertible — lives in the Chronicle integration today, so closing this means Arc.Core growing a key convention of its own. That is a deliberate decision I did not take here; happy to file it as a follow-up.

## Fixes to #2422

- **SP0041 read paging off a query's parameters, which Arc never fills in.** `Paging`/`Sorting` travel on `QueryContext`; `ModelBoundQueryPerformer` treats any non-DI parameter as a query-string argument and throws `MissingArgumentForQuery` for a non-nullable one it cannot bind. What actually marks a query paged is an `IQueryable` return, per `ComputeSupportsPaging` in both performers — so the canonical paged Arc query was silent while the shape that reports is one Arc rejects. Now driven off the return type. The parameter read is kept for a signature that declares the intent.
- **A route was read only from the symbol it was written on, and only the first of them.** ASP.NET Core honors `[Route]` on a base controller and on an overridden action, and `[Route]`/the verb attributes are all `AllowMultiple`. `[Route("api/[controller]")]` on an abstract base is the common pattern, and `IsController` excludes abstract types, so that route was served and never mentioned.
- **The read model `[Path]` report could name a route nothing serves.** A `[Path]` on a query replaces the read model's rather than extending it, so it is now reported only while some query still falls back to it. The spec added in #2422 asserted the wrong thing here.
- Plus `Tags.Of` returning `string[]` instead of double-materializing, consistent whitespace handling between the two `[Path]` reads, and the missing `should_compile_the_source_it_analyzed` guard on the new controller spec.

Not addressed: tags on a reactor/reducer/projection are still dropped silently, and a read model reached only through `IProjectionFor<T>` without `[ReadModel]` never reaches the tag report. `Tags.cs`'s remark promises more than SP0042 delivers — worth a follow-up.

## Fixes to #2421

`verify-links.sh` used one glob per extension, and linkinator aborts on any glob matching nothing — so the `.mdx` glob would fail a repository whose pages are all `.md` while every link in it is fine. One `{md,mdx}` glob now. The scanned-link count was also scraped from output that linkinator colorizes whenever `FORCE_COLOR` is set, which made a real link failure report as "the globs matched nothing"; the count is read after stripping the escape codes (`NO_COLOR` alone does not win against `FORCE_COLOR`).

Not addressed: local lint runs unpinned `npx markdownlint-cli2` while CI pins the action, so the two can still drift; and `.mdx` files are linked-checked but never linted.

## Verification

Clean `dotnet build -c Release` from the repository root: 0 errors, 1 warning — a pre-existing `CS0436` in `TestApps/Chronicle` that reproduces on `main` unchanged.

All 15 spec projects run locally. Notable: Arc.Core.Specs 1576, Chronicle.Specs 560, EntityFrameworkCore.Specs 300, Screenplay.Specs 1165 (15 new), MongoDB.Specs 148 (9 new). `Documentation/verify-markdown.sh` passes with 317 links scanned.

One thing to know about the MongoDB specs: resolving looks up a read model's class map, which resolves the serializer for its id member. If the driver caches its own `Guid` serializer before `AddCratisMongoDB` registers Cratis's, that registration throws — so the spec context initializes the defaults first, the order an application always has.

## Review

Authored by an agent with no human review yet. Please read the ownership design and the `ReportHowItIsServed` rewrite in particular before merging.

woksin:
## Consolidation is complete

#2420, #2421, #2422 and #2423 are closed — every commit on each is contained in `combined/one-release`, verified with `git rev-list --count HEAD..origin/<branch>` returning 0 for all four. Their branches are left in place.

#2424 merged to `main` on its own while this was being prepared, so `origin/main` is merged in here rather than closed.

#2207 is **not** consolidated, and should not be: it is a draft, `CONFLICTING` against `main`, untouched since 15 May, and +363/-1577 across 119 files removing the Castle DynamicProxy and Polly layer. That is a behavioral refactor needing its own review and its own release, not something to fold into a consolidation.

That leaves #2425 and #2207 as the only open pull requests.

## `main` does not currently build

Verified by building `Source/DotNET/Chronicle.Testing` at `origin/main` (`57b72588`): two `CS0535`s, `EventStoreForScenario` missing `IEventStore.ExternalServices` and `IEventStore.Identities`. This branch is the fix.

**`v20.65.4` is a half-published release.** In its publish run ([30515638581](https://github.com/Cratis/Arc/actions/runs/30515638581)):

| Job | Result |
|---|---|
| `release` (tag + GitHub release) | success |
| `publish-npm-packages` | success |
| `publish-dotnet-packages` | **failure** — the two `CS0535`s above |

So the tag, the GitHub release and the npm packages exist at `20.65.4`, and no NuGet package does. NuGet is still on `20.65.3` while npm is on `20.65.4`.

Two consequences worth a decision:

1. **The Chronicle bullets are kept in this description even though `v20.65.4`'s notes already carry them.** No NuGet consumer has received those changes, so the release that actually delivers them should describe them. The cost is that a reader of the GitHub releases page sees them twice. Say the word and I will strip them, or delete the `v20.65.4` release and tag so this one carries the whole story.
2. Closing the four pull requests each triggered a `Publish` run (it fires on `pull_request: closed`). Both were correct: `release` succeeded, `publish-dotnet-packages` and `publish-npm-packages` **skipped**, no tag or release created. `v20.65.4` is still the latest.

## Verification after merging `main`

Clean `dotnet build -c Release` from the repository root: 0 errors, 1 warning — the same pre-existing `CS0436` in `TestApps/Chronicle` that reproduces on `main`.

PR #2430, opened 2026-07-30
Take Screenplay modules from the namespaces when no project names the application

woksin:
# Summary

A solution holding several applications was described as one module named after the solution, with each application a feature of it.

## Changed

- A Screenplay document generated from several projects now declares one module per outermost namespace segment, instead of gathering every project into a single module named after the solution. Naming a module still produces that one module, and generating from a single project is unchanged. (#2399)

### DIFF
--- a/Source/DotNET/Screenplay.Specs/for_ScreenplayGenerator/when_generating/from_projects_in_separate_namespace_roots/SeparateRootsSource.cs
+++ b/Source/DotNET/Screenplay.Specs/for_ScreenplayGenerator/when_generating/from_projects_in_separate_namespace_roots/SeparateRootsSource.cs
@@ -1,0 +1,85 @@
+// Copyright (c) Cratis. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.CodeAnalysis;
+
+namespace Cratis.Arc.Screenplay.for_ScreenplayGenerator.when_generating.from_projects_in_separate_namespace_roots;
+
+/// <summary>
+/// The source of a solution holding two applications that share nothing, the way a samples or a monorepo solution
+/// really is arranged.
+/// </summary>
+/// <remarks>
+/// The layered source is two projects of one application, so every namespace in it begins with the same segment and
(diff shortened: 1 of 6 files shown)


woksin:
Reviewer context.

**Reported from the CLI.** `cratis screenplay generate Samples.slnx` against Cratis/Samples produced `domain Samples` / `module Samples` with `feature Library` and `feature Quickstart` beneath it. This is the `SegmentsToSkip`/module-inference item #2399 flagged as still open after the multi-compilation overload landed.

**Why namespace roots and not assemblies.** `SliceUnion` is explicit that "a slice is recovered from a namespace, and nothing says a namespace belongs to one project" — a contracts project and the project beside it declare one slice from two compilations. A module per assembly would have to place a slice that came from two, so there is no such grouping to make. Namespace roots also match what the reporter expected: in that solution `Quickstart` is not a project, it is four (`Common`, `Common.AspNetCore`, `Console`, `AspNetCore`), and grouping by assembly would have produced four modules named after them.

**The condition** is the one `AnalyzedCompilations.NameOf` already uses to refuse to name the domain after one project — nothing named a module, and no single assembly names the application. `ModulesFromNamespaceRoots` is resolved rather than configured, and sticky: `ScreenplayEmitter` resolves a second time against `model.Domain`, which by then is known, and without stickiness the module would be filled in from it and the document would come back as one module after all.

**Verification**

- `dotnet test` on Screenplay.Specs — 3519 passed, 0 warnings (24 new), on a branch cut from `main`
- The four assertions in `and_nothing_names_a_module` were confirmed to **fail** against the previous behavior, so they pin the change rather than restate it. `and_a_module_is_named` passes either way, which is right — that path is unchanged.
- `LayeredSource` is two projects of one application, so every namespace in it starts with the same segment. That is why no existing spec covered this and why none of them moved; the new source is two projects that name themselves.

**Label** is `minor` on the grounds that the API addition is additive, but generated documents for multi-project solutions do change shape — say the word if you would rather it were `major`.

**Downstream:** the CLI needs no change; `ArcScreenplayGeneration` already leaves `Module` unset for a solution. It picks this up when the package version moves off `20.66.0`.

PR #2432, opened 2026-07-30
Stop pull requests cancelling each other's builds

woksin:
# Summary

Build infrastructure only, with no user-facing change and so nothing for the release notes: open pull requests shared one concurrency group and cancelled each other's builds.

### DIFF
--- a/.github/workflows/dotnet-build.yml
+++ b/.github/workflows/dotnet-build.yml
@@ -1,7 +1,13 @@
 name: .NET Build
 
+# The build runs on pull_request_target, where github.ref is the branch being merged into rather than the pull
+# request - refs/heads/main for every one of them. Grouping on it therefore put every open pull request in one
+# group, and cancel-in-progress made each new one cancel the build of whichever was already running. The pull
+# request that lost was left with a cancelled run and no failing job, which reads as though it is still going
+# rather than as though anything is wrong. So the group is the pull request when there is one, and the ref only
+# for the manual runs that have no pull request to name.
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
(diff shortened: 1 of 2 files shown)


woksin:
Reviewer context.

**Found by being hit by it.** #2430 sat with a cancelled `.NET Build`, no failing job, and checks that read as still running. Re-running it produced the same result, because this pull request's own build cancelled it.

Three consecutive runs of `.NET Build`, two of them killed by the next one to start:

| Run | Branch | Result |
|---|---|---|
| 30579102772 | `feat/modules-from-namespace-roots` (#2430) | cancelled |
| 30579887558 | `feat/arc-core-command-key` | cancelled |
| 30580415574 | `fix/pull-requests-cancelling-each-others-builds` (this one) | holds the group |

`feat/arc-core-command-key` is not mine, so this is not a one-off — any two pull requests open at once means one loses its build.

**Why `github.ref` is wrong here.** On `pull_request_target` it names the base branch, so it is `refs/heads/main` for every pull request rather than anything that distinguishes them. `dotnet-build.yml` already tells the two events apart the same way when it decides which commit to check out (`github.event_name == 'pull_request_target' && github.event.pull_request.head.sha || github.sha`); the concurrency group was just never brought along. CodeQL was never affected because it triggers on `pull_request`, which does get a per-pull-request ref.

**What makes it costly rather than merely wasteful** is the shape of the result. A cancelled run reports no failure, so nothing says anything is wrong - waiting on it waits forever, and merging on it merges on a build that never ran.

**Ordering.** `pull_request_target` reads the workflow from the base branch, so this pull request's own run still uses the version on `main` and can be cancelled the same way. The fix takes effect for runs started after it merges. #2430 needs its build re-run once this lands, and `feat/arc-core-command-key` does too.

**Verification** is by inspection and a YAML parse of both files; there is no way to exercise a concurrency group from a branch, for the reason above.

PR #2437, opened 2026-08-01
Make concurrency: true on a command actually enforce concurrency

woksin:
# Summary

`concurrency: true` on a command declared a guarantee that has never been in force.

## Fixed

- A command scoped with `concurrency: true` now actually participates in the concurrency check. The scope carries an expected sequence number resolved by the configured concurrency strategy, instead of a hardcoded `EventSequenceNumber.Unavailable` that the kernel skipped.
- Declaring `concurrency: true` no longer leaves a command with weaker protection than declaring nothing. The scope it built was not `NotSet`, so it displaced the strategy the event sequence would otherwise have applied.
- The scope is bound to the event source being appended to. A null event source id made the kernel compute the tail across every event source sharing the type.
- A command appending across streams gets a concurrency scope per target event source, instead of one scope carrying a single stream's expected tail applied to all of them.


woksin:
Reviewer context — companion to Cratis/Chronicle#3565, which carries the Chronicle-side half of the same report (CHR-33). This is the half that actually restores the guarantee; that one names the failure state and makes the kernel warn instead of failing open.

**The mechanism, since it spans both repos.** `ConcurrencyScopeBuilder` returned `new ConcurrencyScope(EventSequenceNumber.Unavailable, EventSourceId: null, …)`. Three consequences, each verified in source:

1. `ConcurrencyScope.NotSet` is `EventSequenceNumber.Max`, not `Unavailable`, so the scope was *not* `NotSet` — and `EventSequence.Append` only substitutes the strategy scope when the incoming one is null or `NotSet`. The scope passed straight through.
2. `ConcurrencyScope.ShouldBeValidated` requires `SequenceNumber.IsActualValue`, and `IsActualValue` is `!= Unavailable && != Max`. So `ConcurrencyValidator` returned `None()` on its first line and `EventSequence` appended unconditionally.
3. `EventSourceId: null` converts to the contract flag `false`, which the kernel reads as *"do not scope the tail to this event source"* — so had (1) and (2) been fixed alone, the tail would have been computed across every event source sharing the type. That is why the event source id is set here rather than in a follow-up; the two have to move together.

Net effect before this PR: `concurrency: true` disabled the client's default `OptimisticConcurrencyStrategy` and put nothing in its place. Deleting the attribute turned checking back on.

**What is deliberately unchanged.** A command with no concurrency attribute still returns `null`, so the event sequence keeps applying its configured strategy exactly as before. The Arc docs previously said such appends proceed with no check, which does not match what the client does — I corrected the docs to describe the actual behavior rather than changing the behavior to match the docs. Making un-attributed appends genuinely unchecked would *remove* a check that exists today, which is a different risk class and your call, not a side effect of this fix. Happy to follow up if you want the documented semantics instead.

**On the strategy asymmetry** (pre-existing, not touched here): `OptimisticConcurrencyStrategy.GetScope` computes the expected tail with `GetTailSequenceNumber(eventSourceId)` only, while the kernel's `ConcurrencyValidator` recomputes it narrowed by the scope's stream type, stream id, source type and event types. A narrower kernel-side tail can only be ≤ the client's, so this errs toward missing violations rather than inventing them — but the two ought to agree. Worth a separate issue.

**Verification.** `Arc.slnx` builds clean in Debug and Release; the one `CS0436` warning in `TestApps/Chronicle` is pre-existing (confirmed against a clean tree). Full solution spec run green — 4,052 specs across 14 projects, including 578 in `Cratis.Arc.Chronicle.Specs`.

Three existing specs under `when_handling_with_metadata` had been asserting only that *some* scope reached the append. They now run against a stand-in strategy that behaves like the real one, so they assert against a scope that carries a real tail. `ConcurrencyScopeStrategyStandIn` exists for that reason — an unconfigured NSubstitute strategy hands back no scope at all, which no real strategy does, and a spec asserting on that would have been vacuous.

Not verified: no live race was run. Every claim above is read from Arc and Chronicle source; the specs pin scope construction, not two concurrent appends against a real kernel.

PR #2438, opened 2026-08-01
Update Cratis.Chronicle to 16.13.3 and Cratis.Fundamentals to 7.16.8

woksin:
## Changed

- Updated `Cratis.Chronicle` from 16.10.0 to 16.13.3.
- Updated `Cratis.Fundamentals` and `Cratis.Metrics.Roslyn` to 7.16.8, so a collection absent from a payload deserializes to an empty array rather than `undefined`.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -14,11 +14,11 @@
          advisory (GHSA-v5pm-xwqc-g5wc). Pin to the patched 2.x release. -->
     <PackageVersion Include="Microsoft.OpenApi" Version="2.7.5" />
     <!-- Cratis -->
-    <PackageVersion Include="Cratis.Chronicle" Version="16.10.0" />
-    <PackageVersion Include="Cratis.Chronicle.AspNetCore" Version="16.10.0" />
-    <PackageVersion Include="Cratis.Chronicle.Testing" Version="16.10.0" />
-    <PackageVersion Include="Cratis.Fundamentals" Version="7.16.6" />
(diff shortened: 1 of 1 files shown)


woksin:
Second hop of the chain: Cratis/Fundamentals#1090 → Cratis/Chronicle#3567 → here. This is the hop consumers actually feel, since Arc's generated TypeScript proxies declare collection fields as arrays and 7.16.8 is what makes that declaration true at runtime.

**The Chronicle jump is the part worth reviewing.** It moves three minors, from 16.10.0 to 16.13.3, spanning the constraint-definition widening that made 16.12 a store-compatibility boundary, the failed-partition recovery fix, method-level `[OnceOnly]`, and the kernel half of the concurrency-scope work. Arc builds against all of it.

**Verification.** `Arc.slnx` builds clean in Debug and Release, 40 projects. Full solution spec run green — 5,296 across 15 projects, including `Cratis.Arc.Chronicle.Specs` 578 and `Cratis.Arc.ProxyGenerator.Specs` 1,039. The single `CS0436` warning in `TestApps/Chronicle` is pre-existing and reproduces on a clean tree.

**One sequencing note.** Chronicle#3567 (its own Fundamentals bump) merged but will not cut a release: `publish.yml` triggers only on PRs touching `Source/**` or `Docker/**`, and that one touched `Directory.Packages.props` alone. So 16.13.3 is the current Chronicle and this pins to it. Nothing is lost — Arc pins Fundamentals directly, so it gets 7.16.8 regardless of which Fundamentals the Chronicle package was built against. Chronicle's bump will ride out with its next `Source/**` release.

PR #2442, opened 2026-08-04
Let a consumer declare how a type crosses the wire

woksin:
## Added

- `--type-to-ts` argument and matching `TypeToTsType` MSBuild item declare how a .NET type crosses the wire to TypeScript, and take the place of the built-in mapping for that type.

## Changed

- `@cratis/arc` declares `@cratis/fundamentals` as a peer dependency instead of a versioned dependency. Consumers whose package manager does not install peers automatically must now declare it themselves.

### DIFF
--- a/Source/DotNET/Tools/ProxyGenerator.Build/build/Cratis.Arc.ProxyGenerator.Build.targets
+++ b/Source/DotNET/Tools/ProxyGenerator.Build/build/Cratis.Arc.ProxyGenerator.Build.targets
@@ -37,10 +37,11 @@
             <ExcludeTypes>@(ExcludeType -> '--exclude-type=%(TypeName)', ' ')</ExcludeTypes>
             <ExcludeNamespaces>@(ExcludeNamespace -> '--exclude-namespace=%(Namespace)', ' ')</ExcludeNamespaces>
             <NamespaceRoots>@(NamespaceRoot -> '--namespace-root=%(Namespace)=%(Folder)', ' ')</NamespaceRoots>
+            <TypeMappings>@(TypeToTsType -> '--type-to-ts=%(TypeName)=%(TsType)=%(Package)', ' ')</TypeMappings>
         </PropertyGroup>
 
         <Exec ConsoleToMsBuild="true"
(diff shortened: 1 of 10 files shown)


woksin:
Reviewer context — verification and design notes, kept out of the release notes.

**Design**

- The mapping is consulted **ahead of** `_primitiveTypeMap`. Consulted after, it could only ever add a type the generator has never seen; ahead, it can also correct one it already knows, which is the case the seam exists for. A spec covers the ordering, and reversing it turns three specs red.
- Deliberately an MSBuild item rather than a `[TsType]`-style attribute: a build-time concern belongs in the build file, and an attribute cannot express a mapping for a type the consumer does not own. This follows the `AssemblyToPackageMapping` pattern already in the targets.

**Verification**

- `dotnet build -c Release` 0 warnings 0 errors; `ProxyGenerator.Specs` 1055 specs pass, including all pre-existing ones untouched; `Source/JavaScript/Arc` `yarn ci` 749 specs.
- Generated proxies are committed in consumer repositories, so the guarantee worth protecting is that a build configuring nothing generates exactly what it generated before. A spec covers configure-then-clear leaving no residue, and both mutations (setter stores nothing; lookup moved after the built-in map) were confirmed to turn the intended specs red.

**Notes**

- The peer dependency change is unrelated to the generator feature and is a separate commit. Two copies of `@cratis/fundamentals` in one realm give each its own converter registry and class objects: a converter registered on one reaches only that one, `instanceof` against the other is false, and a version pinned at the top level never reaches a nested copy. Measured: a sibling declaring it as a dependency pinned to 7.16.7, in an app pinning 7.16.8, installs two copies; declared as a peer, one.
- Unrelated to this PR but worth knowing: `DateOnly` and `TimeOnly` still map onto `Date` on `main`. A fix exists on the local `ada-upstream-backlog` branch and has not been pushed.

woksin:
Relabelled **minor** and the peer-dependency change moved out to #2445.

Bundling them was wrong: this half is purely additive and opt-in — a build passing no `--type-to-ts` generates byte-identical output, which is the property that matters since proxies are committed in consumer repositories. The peer-dependency change is the only part that can break an install, and it deserves its own label rather than dragging this one to major.

woksin:
Recombined with the peer-dependency change and relabelled **major**; #2445 is closed as redundant.

`publish.yml` triggers on every merged PR touching `Source/**`, so two PRs here would have cut two Arc releases. One release is preferred, and a release takes the highest bump of its contents — the additive generator feature ships under the major that the dependency change requires, which is the correct semantics for the release even though the feature itself is not breaking.

The two changes remain separate commits, so they stay independently reviewable and revertable.

woksin:

woksin:
Noting for whoever picks this work up, since this PR was closed unmerged and the branch still exists.

CI on this PR failed one job — `proxy-specs (Cratis.Arc.ProxyGenerator.for_TypeExtensions)`. The four specs that configure type mappings race each other: `SetTypeMappings` replaces a static table wholesale, so two specs configuring it concurrently each resolve against the other's table. Distinct subject types do not help, because the replacement takes the whole table with it.

Fixed in `b514f329` on this branch (`feature/proxy-generator-type-mappings`) by putting them in a collection with `DisableParallelization`. Measured on the filtered job CI runs: **9 failures in 10 runs without it, 0 in 10 with it**. Full assembly 1055 pass.

⚠️ Any branch that carries this type-mapping seam **without `b514f329`** will hit the same failure. It reproduces on the filtered `for_TypeExtensions` job and generally does not on a full-assembly run, which is why it reached CI in the first place.

PR #2443, opened 2026-08-04
Identify command form fields and columns by a tamper-resistant marker

woksin:
# Summary

A command form field or column is now identified by a marker a build transform does not touch, with the legacy `displayName` kept as a permanent fallback.

## Added

- `markAsCommandFormField` and `markAsCommandFormColumn` for marking a hand-rolled field or column, and `isCommandFormField` and `isCommandFormColumn` for identifying one, exported from `@cratis/arc.react/commands`
- `CommandFormMarked`, the marker shape, and the `CommandFormFieldDisplayName` and `CommandFormColumnDisplayName` constants
- Documentation on how a field is recognized, on marking a hand-rolled field, and on the Storybook setting that breaks it

## Changed

- `CommandForm` identifies field and column children through the marker, falling back to `displayName`, so consumers marking a field by hand are unaffected
- `asCommandFormField`, `CommandFormField`, `RadioButtonField`, `RadioGroupField` and the column component set the marker alongside the `displayName` they already set

## Fixed

- A command form field whose `displayName` is rewritten by a build transform — such as Storybook's `reactDocgen: 'react-docgen-typescript'`, which rewrites it by default — is no longer silently unbound from its command, rendering an input that accepts typing while nothing reaches the command

### DIFF
--- a/Documentation/frontend/react/command-form/custom-fields.md
+++ b/Documentation/frontend/react/command-form/custom-fields.md
@@ -72,6 +72,24 @@ The second parameter to `asCommandFormField` is a configuration object:
 | `defaultValue` | `TValue` | Yes | Default value when the field is empty/undefined |
 | `extractValue` | `(event: unknown) => TValue` | No | Function to extract the value from change events. If omitted, the event itself is used as the value. |
 
+## How a field is recognized
+
+`asCommandFormField` marks the component it returns so `CommandForm` can pick its fields out of its children. The marker is a static `isCommandFormField` property, and the component's `displayName` is set to `CommandFormField` as well, which is what a version of `@cratis/components` that predates the marker still reads.
+
+:::caution
+**Do not overwrite `displayName` on a command form field**, and do not strip static properties from one. A build transform that does either can unbind the field from its command — the form renders, the input accepts typing, and nothing reaches the command. There is no error and no warning.
+
+The most likely source is Storybook's `reactDocgen: 'react-docgen-typescript'` setting, whose plugin rewrites `displayName` by default. The `isCommandFormField` marker is there to survive exactly that, so a field keeps working under it — but a transform that removes both loses the binding.
+:::
+
+If you build a field by hand rather than through `asCommandFormField`, mark it with `markAsCommandFormField` rather than assigning `displayName` yourself:
+
+```tsx
+import { markAsCommandFormField } from '@cratis/arc.react/commands';
+
+markAsCommandFormField(MyHandRolledField);
+```
+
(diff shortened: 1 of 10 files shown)


woksin:
Reviewer context — none of this is release-note material.

**Pairs with Cratis/Components#112.** That PR carries the consumer half: `CommandDialog` and `CommandStepper` read the field marker written here, and `CommandDialog.Column` writes the column marker read here. Either can ship first — both sides keep the legacy `displayName`, so a new version of one works against an old version of the other in both directions.

**Marker shape.** A static `isCommandFormField` / `isCommandFormColumn` boolean rather than a `Symbol`. A plain property needs no cross-package import, so neither package has to know the other's version — which matters because `@cratis/components` declares this package as a version range, and a named import would be a hard module-link error against any version in that range predating the marker. It is also visible in DevTools and trivial for a consumer to set by hand.

**The two packages briefly disagreed, and every gate stayed green.** This repository chose the boolean while `@cratis/components` independently chose `Symbol.for`. Nothing threw, because both kept the `displayName` fallback — and that is what hid it: the marker simply stopped crossing the package boundary, so a field whose `displayName` a build transform had rewritten still bound in a bare `CommandForm` and silently unbound inside a `CommandDialog`. The failure the marker was added to prevent, surviving the fix, invisible to every spec in both repositories. It was only findable by reading the other repo.

`when_exchanging_marked_components_with_a_consuming_package` is the guard that closes it, asserting both directions: a component marked the way a consuming package marks one, with its `displayName` then overwritten, is recognized here; and a component marked here exposes the marker under the property name that package reads, with the legacy `displayName` still set. Renaming either marker now reds it.

**The `displayName` fallback should not get a deprecation path.** It is what lets the two independently versioned packages interoperate in both directions, and `@cratis/components` has a spec that stamps the string on a fake to make it a field.

**Verification.** `yarn ci` green in `Arc.React` — 157 spec files, 398 tests — with lint clean. Mutation-proven rather than asserted: renaming the marker property reds the cross-package spec. Worth noting that `vitest` does not typecheck, so the specs passed while `tsc -b` in the build step still failed on a bad cast; green tests alone would have shipped a broken build.

**Open for a maintainer:** whether a dev-build `console.warn` for a `CommandForm` that receives children and recognizes zero fields is wanted. It would have ended the original investigation in one page load, but whether this package wants runtime warnings at all is a house-style call, so it is not included here.

No issue references in the description: a search of this repository's issues found nothing matching, and inventing or reusing a number is worse than omitting one. Happy to add them.

woksin:

PR #2444, opened 2026-08-04
Stop the proxy generator and command pipeline erasing distinctions the client needs

woksin:
# Summary

Twelve defects and missing seams found by a consumer, each reproduced here before being fixed. Three are breaking.

## Added

- `ValidationResultReason` on `ValidationResult`, so a rejection can be told apart from another without matching message text
- `ShouldHaveValidationErrorBecauseOf`, for asserting why a command was rejected
- `ICommandResultAssertionPolicy`, a seam for applying a repository's own policy on top of the built-in `CommandResult` assertions
- `ARC0015`, warning when a query parameter is declared raw and converted to a concept in the method body, skipping the concept's validator
- `markAsCommandFormField` and `markAsCommandFormColumn`, so a command form field survives a build transform that rewrites `displayName`
- Documentation for how .NET types map to TypeScript in generated proxies

## Changed

- `DateOnly` and `TimeOnly` now generate as `string` rather than `Date`, which rendered a calendar date a day early west of UTC and a time of day as `Invalid Date`
- A validator message declared as a factory is no longer resolved at generation time and baked into the generated client validator; the rule still mirrors, the message is left to the server
- `autoServerValidateThrottle` now governs the per-keystroke validation request instead of a second, redundant one
- `ShouldHaveValidationErrors` no longer passes when the only thing that rejected the command was a dependency the pipeline could not resolve

## Fixed

- Cross-references in XML documentation are rendered into the generated TypeScript instead of erased, along with the prose that fused around them
- Documentation stating that routing-only stream metadata does not affect concurrency control, which it does
- Documentation not saying that the provider resolving a command-side read model also decides which serialization boundary it crosses


woksin:
Reviewer context — not release notes.

## How each change was arrived at

Every item was reproduced in this repository before being fixed, and mutation-proven: red before, green after, red again with the fix deleted. Three are worth a reviewer's attention because the evidence is unusual.

**`autoServerValidateThrottle` was measured, not reasoned about.** Ten characters typed into one field, counting POSTs to `{route}/validate`: 11 requests at the default throttle, 10 with the throttle raised past the burst — the prop moved nothing. Now 1 and 0. The eleventh request is a second, throttled call that only exists because the first one already ran. `when_typing_into_a_field_with_autoServerValidate`.

**That fix makes the code match documentation it had diverged from.** `Documentation/frontend/react/command-form/auto-server-validation.md` already promised *"With throttle: 1 server call (after user stops typing for 500ms)"*. The page needed no change.

**The routing-only concurrency claim is now pinned end to end.** `without_concurrency_on_event_source_type` asserts the pair that matters — a routing-only command reaches `Append` with **no scope** and **its tag intact** — which is what makes the fallback strategy narrow by it. Asserting the null scope alone is what let the misconception survive; the existing `ConcurrencyScopeBuilder` spec's rationale said so in source and is corrected here.

## Design decisions a maintainer may want to revisit

- **`ValidationResultReason` is an open `ConceptAs<string>`, not an enum.** Rejections are composed in Arc, in Chronicle and in application code, so a closed set makes every new kind a breaking change for anyone switching over it. It is deliberately *not* carried in `ValidationResult.State`, which already carries FluentValidation's `WithState` and therefore belongs to whoever wrote the rule.
- **FluentValidation's `ErrorCode` is still not carried through.** One line, but it would make that library's defaults (`NotEmptyValidator`, …) a wire-visible value Arc then owes.
- **`ARC0015` is a `Warning`, and covers internal as well as public query methods.** The shape is legal and retyping is not behaviour-neutral — an omitted argument deserializes to `null` rather than `NotSet`, and a strict validator becomes reachable — so an `Error` would be suppressed rather than adopted. Both caveats are in the descriptor's `description`. It registers an operation-block action, which has no precedent in either analyzer project here.
- **`DateOnly` → `string` is a floor, not a ceiling.** A first-class `DateOnly` in `@cratis/fundamentals` would be better; that package ships no such type today, and an Arc-defined class cannot work because the converter registry is module-private. If Fundamentals ships one, this becomes a two-line change.
- **The command form marker is a static property, not a `Symbol`.** A `Symbol` resists a name-based transform more thoroughly but is a public-surface choice, and needs `Symbol.for` to survive a duplicate install. The `displayName` fallback stays permanently on both sides — it is what lets independently versioned packages interoperate in both directions. The Components half is Cratis/Components#112.

## Verification

- `dotnet build` Debug and Release: clean. Release multi-targets and treats warnings as errors.
- 15 C# spec projects green; `Cratis.Arc.Testing` had no spec project and no assertion coverage before this, and now has both.
- 5 TypeScript workspaces green across three consecutive runs; `yarn lint` and `npx tsc -b` clean on each.
- Pre-existing and unrelated: a `CS0436` warning in `TestApps/Chronicle`, reproduced on a clean tree with `--no-incremental`; and `for_ObservableQueryDemultiplexer…and_query_context_has_paging`, which flakes under parallel load and passed 50/50 on three isolated runs.

## Coverage gaps this closed on the way

`ValidatorInvoker`, the `Cratis.Arc.Testing` assertions, and the aggregate-root concurrency path each had no direct coverage at all — which is how the aggregate-root path came to carry the same flattening as the event-log path unnoticed. An existing throttle spec asserted `toBeLessThanOrEqual(5)`, which is what let a round trip per keystroke hide underneath it; it now asserts `toBe(1)`.

woksin:

PR #2445, opened 2026-08-04
Depend on @cratis/fundamentals as a peer

woksin:
## Changed

- `@cratis/arc` declares `@cratis/fundamentals` as a peer dependency instead of a versioned dependency. Consumers whose package manager does not install peers automatically must now declare it themselves.

### DIFF
--- a/Source/JavaScript/Arc/package.json
+++ b/Source/JavaScript/Arc/package.json
@@ -69,7 +69,12 @@
         "up": "yarn g:up"
     },
     "dependencies": {
-        "@cratis/fundamentals": "^7.16.0",
         "rxjs": "^7.8.2"
+    },
+    "peerDependencies": {
+        "@cratis/fundamentals": "^7"
+    },
+    "devDependencies": {
+        "@cratis/fundamentals": "7.16.0"
     }
 }


woksin:
Split out of #2442, where it was wrongly bundled with an additive generator feature.

**Why major.** `@cratis/arc` no longer brings `@cratis/fundamentals` transitively. npm 7+ installs peers automatically and is unaffected; Yarn Berry and pnpm without `auto-install-peers` are not, and a consumer there must declare it.

**Why do it.** Two copies in one realm is a broken outcome, not a wasteful one — each gets its own converter registry and class objects, so a converter registered on one reaches only that one, `instanceof` against the other is false, and a version pinned at the top level never reaches the nested copy. Measured against real published versions: a sibling declaring it as a dependency pinned to 7.16.7, in an app pinning 7.16.8, installs two copies; declared as a peer, one.

The `^7` range is deliberately wide so a patch or minor release does not force lockstep releases, and an exact `devDependencies` entry keeps this repository's own build pinned.

woksin:
Closing as redundant — recombined into #2442.

Splitting this out gave each change an accurate label, but `publish.yml` cuts a release per merged PR touching `Source/**`, so it would also have produced two Arc releases. One release is preferred, and #2442 now carries both changes labelled `major`, which is the correct bump for a release containing this one.

woksin:

PR #2446, opened 2026-08-04
Harden the proxy generator, the validation surface and the command form marker

woksin:
# Summary

One release covering the proxy generator, the command pipeline's validation surface, and the command form. Supersedes #2442, #2443, #2444 and #2445. Three changes are breaking.

## Added

- `TypeToTsType` build item, declaring how a .NET type crosses the wire — consulted ahead of the built-in map, so it corrects an existing type as readily as it declares a new one
- `ValidationResultReason` on `ValidationResult`, so a rejection can be told apart from another without matching message text, which is presentation and changes without notice
- `ShouldHaveValidationErrorBecauseOf`, for asserting why a command was rejected — asserting only that it failed passes just as readily when it failed for the wrong reason
- `ICommandResultAssertionPolicy`, a seam for applying a repository's own policy on top of the built-in `CommandResult` assertions, so a shared convention lives in one place instead of every spec
- `ARC0015`, warning when a query parameter is declared raw and converted to a concept in the method body, which skips the concept's validator so an invalid value reaches the handler as though it had been validated
- `markAsCommandFormField` and `markAsCommandFormColumn`, so a command form field survives a build transform that rewrites `displayName` — Storybook's `reactDocgen: 'react-docgen-typescript'` does that by default, which unbound every field with nothing in the output to say why
- `beginSilentValidation` on the command form context, for a custom field to claim a token before it validates so a result a later run has overtaken is discarded rather than applied
- Documentation for how .NET types map to TypeScript, and for declaring your own

## Changed

- `DateOnly` and `TimeOnly` now generate as the `DateOnly` and `TimeOnly` types from `@cratis/fundamentals` rather than as `Date`, which rendered a calendar date a day early west of UTC and a time of day as `Invalid Date`. Each has a converter of its own, so the value arrives as a type rather than as text every call site has to parse. Requires `@cratis/fundamentals` 7.17.0
- A validator message declared as a factory is no longer resolved at generation time and baked into the generated client validator, because a factory can depend on state that does not exist then, making the baked result one arbitrary evaluation; the rule still mirrors, the message is left to the server
- `autoServerValidateThrottle` now governs the per-keystroke validation request instead of a second, redundant one, so typing no longer issues an unthrottled request per keystroke
- `setSilentValidationResult` on the command form context takes the token from `beginSilentValidation` and returns whether the result was applied; called without one it behaves as it did
- `ShouldHaveValidationErrors` no longer passes when the only thing that rejected the command was a dependency the pipeline could not resolve, which let a spec report success against a pipeline that never ran
- `@cratis/fundamentals` is declared as an ordinary dependency by both `@cratis/arc` and `@cratis/arc.react`, at the same range, so an application resolves a single copy; two copies each carry their own `Guid` and converter registry and neither recognizes the other's types. The range is `^7.17.0`
- `Cratis.Fundamentals` is referenced at 7.17.0, level with the npm packages

## Fixed

- A `CommandForm` no longer lets a validation result that a later run has overtaken decide `isValid` — overlapping runs resolving out of order could leave the submit button greyed out for good, with every field valid and no message shown anywhere
- `@cratis/arc.react` declares its own dependency on `@cratis/fundamentals`, which it imports across 20 files but declared nowhere — it resolved only by hoisting through `@cratis/arc`
- Cross-references in XML documentation are rendered into the generated TypeScript instead of erased, along with the prose that fused around them
- Documentation stating that routing-only stream metadata does not affect concurrency control, which it does
- Documentation not saying that the provider resolving a command-side read model also decides which serialization boundary it crosses


woksin:
Reviewer context — not release notes.

## Why one PR

#2442, #2443, #2444 and #2445 are all in here, so this lands as one release rather than four. Two of them overlapped and could not have been merged independently as they stood:

- **#2443 was fully contained in #2444.** Diffing the two branches over the CommandForm paths, the marker files were byte-identical; the only delta was #2444's `autoServerValidateThrottle` work on top.
- **#2442 and #2444 both changed `TypeExtensions.cs`.** `git merge-tree` reported `changed in both`. The real merge resolved it — the two edits are adjacent, not overlapping — but the *design* question underneath needed an answer, below.

## The decision this PR makes

`SetTypeMappings` (#2442) lets a consumer declare how a type crosses the wire. Correcting `DateOnly`/`TimeOnly` (#2444) changes what the built-in map does by default. Those are two answers to one question, and this PR takes both, in that order of precedence:

**The built-in default is correct without configuration; the seam is for wanting something better.** A calendar date crossing as its ISO-8601 string loses nothing and cannot be rendered a day early. A consumer who wants a real `LocalDate` type maps it. The alternative — ship the seam, leave the default wrong — makes every consumer discover a timezone bug before configuring their way out of it, which is exactly how it shipped in the first place.

`when_a_type_mapping_replaces_the_temporal_default` pins the composition, which neither PR specified because neither owned both halves: the defaults are `string`, an unmapped sibling stays `string`, and a mapping replaces one and imports it from its package.

## How the changes were arrived at

Each was reproduced here before being fixed and mutation-proven — red before, green after, red again with the fix deleted. Three are worth attention because the evidence is unusual.

**`autoServerValidateThrottle` was measured.** Ten characters typed into one field, counting POSTs to `{route}/validate`: 11 requests at the default throttle, 10 with the throttle raised past the burst — the prop moved nothing. Now 1 and 0.

**That fix makes the code match documentation it had diverged from.** `auto-server-validation.md` already promised *"With throttle: 1 server call (after user stops typing for 500ms)"*. The page needed no change.

**The routing-only concurrency claim is pinned end to end.** `without_concurrency_on_event_source_type` asserts the pair that matters — a routing-only command reaches `Append` with **no scope** and **its tag intact**. Asserting the null scope alone is what let the misconception survive; the `ConcurrencyScopeBuilder` spec's own rationale said so in source and is corrected here.

## Design decisions a maintainer may want to revisit

- **`ValidationResultReason` is an open `ConceptAs<string>`, not an enum.** Rejections are composed in Arc, in Chronicle and in application code, so a closed set makes every new kind a breaking change for anyone switching over it. Deliberately not carried in `ValidationResult.State`, which already carries FluentValidation's `WithState` and belongs to whoever wrote the rule.
- **FluentValidation's `ErrorCode` is still not carried through.** One line, but it would make that library's defaults (`NotEmptyValidator`, …) a wire-visible value Arc then owes.
- **`ARC0015` is a `Warning`, and covers internal as well as public query methods.** The shape is legal and retyping is not behaviour-neutral, so an `Error` would be suppressed rather than adopted. It registers an operation-block action, which has no precedent in either analyzer project here.
- **The command form marker is a static property, not a `Symbol`.** The `displayName` fallback stays permanently on both sides — it is what lets independently versioned packages interoperate in both directions.

## Verification

`dotnet build` Debug and Release clean; 15 C# spec projects green; 5 TypeScript workspaces green; `yarn lint` and `npx tsc -b` clean.

Pre-existing and unrelated: a `CS0436` warning in `TestApps/Chronicle`, reproduced on a clean tree with `--no-incremental`; and `for_ObservableQueryDemultiplexer…and_query_context_has_paging`, which flakes under parallel load and passed 50/50 on three isolated runs.

## Must ship together

Cratis/Components#112 carries the other half of the command form marker. A version of Components that only knows `displayName` paired with an Arc that only sets the marker would silently unbind every field in a `CommandDialog` — the exact failure the marker prevents. The `displayName` fallback on both sides is what makes the two interoperate in either direction, which is why it is not on a deprecation path.

## Coverage gaps closed on the way

`ValidatorInvoker`, the `Cratis.Arc.Testing` assertions, and the aggregate-root concurrency path each had no direct coverage — which is how the aggregate-root path came to carry the same flattening as the event-log path unnoticed. An existing throttle spec asserted `toBeLessThanOrEqual(5)`, which let a round trip per keystroke hide underneath it; it now asserts `toBe(1)`.

woksin:
Two additions since the branch was opened.

**A flaky spec, caught and fixed.** `when_a_type_mapping_is_configured` and `when_a_type_mapping_is_configured_without_a_package` both map `System.Version`, and `SetTypeMappings` replaces a static table wholesale — so with xUnit running spec classes in parallel they read each other's table. Five consecutive runs of unchanged code failed 3, 3, 2, 2 and 1. Clearing the table in teardown cannot fix it: the race is between one class's setup and another's assertion, both inside their own lifetimes. Serializing them into a non-parallel collection does; six runs clean.

Worth knowing this was live before the merge, not introduced by it — it would have flaked in CI intermittently and looked like an unrelated infrastructure blip.

**The temporal mapping is recorded as provisional.** `DateOnly`/`TimeOnly` → `string` is correct — the string is exactly what the server sent — but it is not descriptive, and it is the one change in this PR that reduces type information rather than adding it. A first-class `DateOnly` in `@cratis/fundamentals`, carrying converters the way `Guid` and `TimeSpan` do, is the better answer.

That cannot ride in this PR: it is a different package with its own release, and Arc cannot name a type that is not published yet — a generated proxy would reference an export that does not exist. So the intent is recorded in the map's own comment and on the documentation page instead of being left to look settled. The wire payload is identical either way, so moving to it later is not a breaking change for anyone reading the value.

Suggested follow-up, after this releases: add `DateOnly`/`TimeOnly` with converters to `@cratis/fundamentals` — now unblocked by `JsonSerializer.registerConverter` — then flip the two map entries here to `FromPackage`. Nothing else in the generator changes.

woksin:
Reviewer context for the two dependency commits (not release-note material):

**Verified locally** — `tsc -b` clean for both `Arc` and `Arc.React`; `vitest run` 400 passed / 158 files. Arc gitignores `yarn.lock` (`.gitignore:300`), so the manifests are the whole change.

**How the missing declaration was found** — `@cratis/arc.react` imports `Constructor`, `field`, `Guid` and `JsonSerializer` from `@cratis/fundamentals` in 20 files (`Bindings.ts`, `identity/IdentityProvider.tsx`, the query hooks) with no declaration in any dependency block. `react-dom` was checked for the same pattern and is clean — devDependency only, never imported from source.

**Worth a follow-up, not fixed here** — the exact `7.16.0` devDependency pins nest a private copy of fundamentals under each workspace while the root hoists 7.16.8:

```
node_modules/@cratis/fundamentals                      7.16.8
Source/JavaScript/Arc/node_modules/@cratis/fundamentals        7.16.0
Source/JavaScript/Arc.React/node_modules/@cratis/fundamentals  7.16.0
```

The `Arc/` copy predates this PR; I matched the existing pin on `Arc.React` rather than diverge. But that is two copies in one realm, which is exactly what `duplicateInstanceGuard` in Cratis/Fundamentals#1092 is built to detect — once that ships, local builds and Storybook will start reporting it. Deduping to `^7.16.0` devDependencies and verifying the range floor in a separate CI job instead would avoid both.

einari:
@woksin I see there is a breaking change here with regards to making DateOnly and TimeOnly a string.
My suggestion is to instead introduce DateOnly and TimeOnly in Fundamentals.

Another thing; making @cratis/fundamentals a peer dependency. Not sure I understood the motivation for this. I would argue Fundamentals is not something one deliberately takes a dependency to, as one is not actually looking for its functionality in an application - but rather one is looking for what Arc can do and typically end up having @cratis/arc.react - which then has @cratis/arc, which then har @cratis/fundamentals as dependency.

woksin:
@einari Agreed on both counts being worth stopping for.

**`DateOnly`/`TimeOnly` in Fundamentals** — you're right, and it's up: Cratis/Fundamentals#1093. Value types with `parse`/`toString` over the ISO-8601 form the server sends, plus registered converters, following the shape `Guid` and `TimeSpan` already have.

Once that releases, the two entries here become `FromPackage` entries pointing at it and nothing else in the generator changes. The wire payload is identical — `"2026-05-12"` either way — so moving from the string to the type is not a second breaking change for anyone reading the value.

The consequence for this PR is that it should wait for that release rather than ship the string mapping. The alternative is dropping the temporal change from here and doing it in a later Arc release, which costs a second release for the same fix. I'd rather hold this one; nothing else in it is blocked.

Two things in #1093 worth your eye, since they're judgement calls rather than mechanics: `DateOnly.toDate()` is a method rather than what the value is, because constructing a `Date` invents a time nobody sent — so the caller who needs one makes that choice visibly. And `TimeOnly.parse` truncates the fraction to milliseconds rather than rounding, so a parse followed by a render never moves the value forward.

**The peer dependency** — your reasoning about intent is right: nobody sets out to depend on Fundamentals, they want what Arc does, and the chain `@cratis/arc.react` → `@cratis/arc` → `@cratis/fundamentals` is how it should read from the outside.

The motivation was a different one, and I should have stated it here rather than leaving it to be inferred. Fundamentals' type registry keys on **constructor identity** — `typeConverters` and `typeSerializers` are `Map<Constructor, …>`. Two copies of the package resolved into one tree therefore fail silently rather than loudly: a converter registered on one copy is invisible to the other, and an `instanceof` against the wrong copy is false while everything still compiles and nothing throws. `duplicateInstanceGuard.ts` exists in Fundamentals precisely because that failure has no other symptom. A peer dependency is the structural prevention; a normal dependency is what permits the duplicate to be installed in the first place.

That said, it's a hazard a lockfile and dedupe usually handle, and the DX cost you're pointing at is real and immediate where the hazard is occasional. Happy to revert it to a normal dependency and let the guard carry that risk instead — your call, and I'd rather take your read on it than trade a concrete papercut for a rare one.

einari:
@woksin Looks great in Fundamentals :) Nice having things typed.

woksin:
@einari Both of your points are addressed in `bdc40f7c`.

**`DateOnly`/`TimeOnly`** now map to the types Fundamentals 7.17.0 ships, imported from `@cratis/fundamentals` exactly the way `Guid` and `TimeSpan` are. The string mapping is gone. The wire payload is unchanged, so this is the same value arriving as a type instead of as text.

**The peer dependency is reverted** to an ordinary dependency. Your reasoning holds, and there is a stronger argument for it than the one I gave against it: a peer does not reach the consumer transitively through `@cratis/arc.react` at all — `arc.react` never declares Fundamentals — so with Yarn it surfaces as a `YN0002` "doesn't provide" warning that the application silences by declaring a package it never imports. The `react` peer on `arc.react` is the pattern used correctly; Fundamentals is not that kind of thing.

It also inverts what it was meant to prevent. Duplicates matter because the converter registry keys on constructor identity — but with a plain transitive dependency and no application-level declaration there is exactly one range in the tree. Making the consumer declare a version is the only way a second, incompatible one gets in.

**The duplicate-install failure is not hypothetical, and this ran into it.** The workspace was carrying `^7.16.2`, `^7.16.0` and `^7.17.0` at once, which resolved two copies — 7.16.8 at the root and 7.17.0 under `Source/JavaScript/Arc`. The scenario specs run in V8 and resolve `node_modules` by walking up, so they got the older copy, which has no `DateOnly`, and failed with `TypeError: targetType is not a constructor`. Aligning all three ranges resolved it to one copy and the specs pass.

That is worth noting for the peer-versus-dependency question rather than against it: the copies came from Cratis's own workspace disagreeing with itself, not from a consumer. Keeping the ranges aligned across our packages is the control that actually holds, and a CI check for it would catch the real cause. Happy to add one if you want it.

The manifest spec that pinned the peer range is kept and retargeted at the dependency — the invariant it encodes (the range's floor must equal the version actually built against) is worth as much either way.

One piece of commit hygiene I got wrong and cannot fix without a force-push: both changes landed in `bdc40f7c`, whose message describes only the temporal mapping. The dependency reasoning is here rather than in the log.

woksin:
Reviewer context for the three commits added on top — the rationale behind two choices that are not obvious from the diff.

### The validation ordering guard sits at the write point, not on `validateSilently`

`validateSilently` is the path that decides `isValid`, so it looks like the one to guard. It is not the only one that writes that state — three do, and they race **each other**, not only themselves:

- `validateSilently` in `CommandForm.tsx` (init, and every `currentValues` change)
- the throttled server round trip in the same file
- the per-keystroke and blur runs in `CommandFormFields.tsx`

Guarding only the first leaves the defect reachable through the other two, so the guard is one token check inside the single writer that all three now go through.

### It discards rather than cancels, deliberately

An `AbortController` per run would also stop a stale result landing, and would save the wasted round trips — but it answers the wrong question. What has to win is the **newest run**, which is not always the fastest one. Aborting in-flight work gets the common case right and drops a legitimately-slow newest verdict on the floor.

So each run claims a token before it starts and hands it back with its result; a result that a later run has already applied is discarded, and nothing is cancelled. `when_a_silent_validation_is_overtaken` asserts both directions for that reason — a guard that simply preferred whatever arrived first would pass the first case and fail the second. Both cases also assert that two runs were genuinely in flight together, or the interesting assertion would hold for a reason unrelated to ordering.

The guard was mutation-checked: reverting the token comparison alone turns the overtaken case red.

### Public surface

`CommandFormContextValue` gains `beginSilentValidation`, and `setSilentValidationResult` takes an optional token and returns whether it applied. Calling it without a token behaves as it did, so a custom field that writes through the context is unaffected — only code that *constructs* the context itself needs the new member.

### The fundamentals manifest spec

`@cratis/arc.react` declared fundamentals as a peer while `@cratis/arc` had it as an ordinary dependency, so the two packages that ship together disagreed about whose job it is to name a version. Both now declare it the same way, and the new spec pins the two ranges **to each other** rather than only to each package's own pin — that is the assertion that catches them drifting apart, which is how an application ends up resolving two copies.

The rationale comment on the existing Arc spec justified not-a-peer with "the application never imports it". That is not true — the generator emits `import { Guid } from '@cratis/fundamentals'` straight into application code, as its own specs assert. The conclusion is unchanged and the comment now says why it holds.

### Verified locally

Release build 41 projects, 0 errors. .NET specs 5,017 passed / 0 failed across all 10 assemblies. `@cratis/arc.react` 409 tests, `@cratis/arc` 756, `tsc` clean, eslint 0 errors.

PR #2455, opened 2026-08-06
Drop the empty assetsDir override from the TestApp Vite configs

woksin:
# Summary

Internal test applications only — no user-facing change.

### DIFF
--- a/TestApps/ArcCore/.frontend/vite.config.minified.ts
+++ b/TestApps/ArcCore/.frontend/vite.config.minified.ts
@@ -16,7 +16,6 @@ import baseConfig from './vite.config';
 export default mergeConfig(baseConfig, defineConfig({
     build: {
         outDir: 'dist-minified',
-        assetsDir: '',
         minify: 'esbuild',
     },
     esbuild: {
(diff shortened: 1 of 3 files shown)


woksin:
Reviewer context.

All three TestApp configs pinned `build.assetsDir` to `''`, flattening every hashed artifact into the `wwwroot` root — one-segment paths sharing no prefix, so a proxy, CDN or WAF path rule has nothing to match on. Removing the line restores Vite's own default. The equivalent fix has already landed in the Templates repo; this brings the TestApps in line with it.

Nothing depended on the flat layout: neither `index.html` carries hardcoded asset paths, neither `Program.cs` calls `UseStaticFiles`, there is no service worker, manifest, CSP or Dockerfile. The one real consumer is the MAUI harness, and `Maui.csproj` links with `%(RecursiveDir)`, which preserves the new nesting. Arc's own `StaticFilesMiddleware` resolves subdirectories and has a spec for it.

ArcCore builds verified — `yarn build` and `yarn build:minified` both emit `assets/` with `index.html` rewritten to `/assets/…`.

`TestApps/AspNetCore` has a pre-existing failure unrelated to this change, confirmed on a clean tree before editing: `tsc -b` fails on `Features/MongoWatcher/MongoWatcher.ts` with `TS2300: Duplicate identifier 'WatchAll'` / `'ObserveAll'` from duplicated generated proxy declarations. Vite never runs, so I verified that stage separately — layout is correct.

Worth noting these TestApp frontends are never built in CI: Maui is absent from `Arc.slnx`, and `run-task-on-workspaces.js` skips `private: true` workspaces. So CI green here does not exercise the change; the verification above is what covers it.

PR #2457, opened 2026-08-06
Materialize read model collections as empty rather than null

woksin:
## Fixed

- A read model's non-nullable collection property reads back as empty rather than `null` when queried through `IMongoCollection<T>`, whether the stored document omits the field or stores `null`

### DIFF
--- a/Documentation/backend/mongodb/convention-packs.md
+++ b/Documentation/backend/mongodb/convention-packs.md
@@ -41,6 +41,93 @@ RegisterConventionAsPack(
 );
 ```
 
+### Read Model Collections Never Null
+
+Materializes a `[ReadModel]`'s declared **non-nullable** collection members as empty collections instead of `null`:
+
+```csharp
+// Registered automatically with name: "Read model collections never null convention",
+// scoped to types marked with [ReadModel]
+RegisterConventionAsPack(
+    conventionPackFilters,
+    ConventionPacks.ReadModelCollectionsNeverNull,
+    new ReadModelCollectionsNeverNullConvention(),
+    type => type.IsReadModel()
+);
+```
+
+#### Why it exists — the driver is a separate boundary from Chronicle
+
+A read model that declares a child collection as a non-nullable `IEnumerable<T>` promises the type system that the
+value is there. The store is free to disagree. Chronicle's read model sink writes **no field at all** for a child
+collection that has never had an element, and can write an **explicit `null`** for one whose last element went away.
+Nullable reference analysis has already concluded the member can never be `null`, so nothing warns, and an unguarded
+`.Any()` / `.All()` / `.Select()` throws at runtime.
+
+Chronicle closes this for **its own reader** with a `JsonTypeInfo` modifier on its client's `JsonSerializerOptions`.
+That fix lives at Chronicle's serialization boundary and **does not reach the MongoDB driver**. Reading a read model
+through `IMongoCollection<T>` — the sanctioned way to query by anything other than the key, and what Arc's own
+server-side paging runs on — goes through the driver instead. This convention is the same guarantee, restated where
+the driver can honor it.
+
+#### What it covers
+
+Both shapes a store can leave behind, through two different mechanisms:
+
+| Stored shape | Mechanism | Result for a non-nullable member |
+|---|---|---|
+| The field is **absent** from the document | a default value on the member map | an empty collection |
+| The field is present and holds **`null`** | a serializer wrapping the member's own | an empty collection |
+
+Both halves are needed: the driver never invokes a member's serializer for an element that is not in the document, and
+a default value is never consulted for one that is.
+
+#### What it deliberately leaves alone
+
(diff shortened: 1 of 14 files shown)


woksin:
Reviewer context.

Closes the last open item from an upstream consumer report. Chronicle's read-model sink omits an empty child collection, and a read model declares that collection as a non-nullable `IEnumerable<T>` — so nullable-reference analysis never warns and an unguarded `.Any()` throws. That produced a live user-facing 400 in a consumer application.

Chronicle already fixed this **for its own reader** (`ReadModelJsonSerialization.WithDeclaredCollectionsNeverNull()`, a `JsonTypeInfo` modifier on the client's `JsonSerializerOptions`). That does not reach the MongoDB driver — and the driver path is the sanctioned one for any lookup by something other than the key, since `IReadModels` has no queryable surface by design. Arc registers `IMongoCollection<T>` as an open generic, Arc's own paging is built on `AsQueryable()`, and Arc's docs show a query taking `IMongoCollection<T>` as the intended shape. So this belongs in Arc: Chronicle's .NET client has no MongoDB dependency at all and should not gain one for this.

**Two stored shapes, two mechanisms — and the second one was nearly missed.** The plan was a `SetDefaultValue` convention, which covers an absent element. Before building it, we checked whether the sink can also write an explicit `null` into a children path. It can: `ChangesetConverter.BuildPropertiesChanged` does `Set(property, converter.ToBsonValue(difference.Changed, path))`, and `MongoDBConverter.ToBsonValue` opens with `if (input is null) return BsonNull.Value`. A `PropertyDifference` with a collection path and a null `Changed` is reachable from `ReducerPipeline.Handle` via `ObjectComparer.CompareValues` — an `IReducerFor<T>` returning a state that omits a collection the stored document had. `SetDefaultValue` does not fire for a present-but-null element, so the fix would have silently covered only half the problem. The convention therefore sets a default-value factory **and** wraps the member serializer.

That null-write path is a code-reading trace with a named call chain, not an executed repro, and it remains a Chronicle-side question — this change tolerates the shape on the reader, it does not stop the sink producing it.

**Scoping.** Applied only to types where `type.IsReadModel()` — the same predicate `MongoDBReadModelForCommandResolver` already uses — and only to members whose declaration is non-nullable. A nullable collection keeps the "no collection" vs "empty" distinction, matching the carve-out Chronicle's own reader makes. Supported shapes are an explicit allow list (`IEnumerable<T>`, `ICollection<T>`, `IList<T>`, `IReadOnlyCollection<T>`, `IReadOnlyList<T>`, `List<T>`, `ISet<T>`, `IReadOnlySet<T>`, `HashSet<T>`, `T[]`) rather than "anything enumerable", so an unconstructible shape is left alone rather than guessed at. `string` and every dictionary shape are excluded.

Two opt-outs, both pre-existing mechanisms: the pack is **named**, so `[IgnoreConventions(...)]` works, and `[BsonDefaultValue]` on a member skips **both** halves so the escape hatch cannot leak through the serializer wrap.

**Overwrite hazard, consciously accepted.** When the element is absent the default replaces whatever an initializer set, so a member initialized to a meaningful non-empty default would be silently emptied. Contained by three things: the `[ReadModel]` scope, the existing house rule that read model properties carry no default values, and the non-nullable-only restriction. A `= []` initializer still yields empty, though as `List<T>` rather than `T[]` — not observable through the declared interface type.

**Checked because the serializer wrap put them at risk**, neither covered by the original investigation:
- LINQ translation still works — `.Where(_ => _.Kids.Any(...))` → `$elemMatch`, `.Select(_ => _.Tags.Count)` → `$size`, skip/take → `$skip`/`$limit`. The wrapper implements `IBsonArraySerializer` and forwards `TryGetItemSerializationInfo`; without that, Arc's `AsQueryable()` paging would have broken silently.
- No class-map re-entrancy for self-referencing (`record Node(…, IEnumerable<Node> Children)`) or mutually-referencing read models.

**A flake introduced and fixed during the work**, worth knowing about: `DatabaseExtensions.NamingPolicy` is process-global, and several specs in this assembly set it to a substitute answering `null` for every name. Once any spec calls `AddCratisMongoDB`, a class map built afterwards renames every element to null. Fixed with `[IgnoreConventions(NamingPolicyNameConvention.ConventionName)]` on the fixtures.

**Verification.** `dotnet build Arc.slnx` Debug and Release, `--no-incremental`: 41 projects, 0 errors, 1 warning — a `CS0436` in `TestApps/Chronicle/Program.cs`, identical at baseline with the change stashed. `dotnet test Arc.slnx -c Debug`: 5461 passed across 16 spec assemblies. MongoDB.Specs: 192 passed, run repeatedly to confirm the flake fix holds.

Two pre-existing flakes seen and baselined, both unrelated to serialization: `for_MongoDBReadModelForCommandResolver…should_resolve_the_document` (1 of 11 baseline runs) and `for_ObservableQueryDemultiplexer…and_query_context_has_paging` in Arc.Core.Specs, which has no MongoDB dependency at all.

Not covered, deliberately: collection members on child element types (children are not `[ReadModel]`), shapes outside the allow list, and multi-dimensional arrays.

PR #2462, opened 2026-08-06
Update Cratis.Screenplay to 1.8.0

woksin:
# Summary

Moves the Screenplay language package to the current release, so the end-to-end check verifies generated documents against the language they will actually be read with.

## Changed

- `Cratis.Screenplay` 1.5.2 to 1.8.0

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -19,7 +19,7 @@
     <PackageVersion Include="Cratis.Chronicle.Testing" Version="16.16.0" />
     <PackageVersion Include="Cratis.Fundamentals" Version="7.17.1" />
     <PackageVersion Include="Cratis.Metrics.Roslyn" Version="7.17.1" />
-    <PackageVersion Include="Cratis.Screenplay" Version="1.5.2" />
+    <PackageVersion Include="Cratis.Screenplay" Version="1.8.0" />
     <!-- MAUI -->
     <PackageVersion Include="Microsoft.Maui.Controls" Version="10.0.90" />
     <PackageVersion Include="Microsoft.Extensions.Logging.Debug" Version="10.0.10" />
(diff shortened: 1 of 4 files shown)


woksin:
Reviewer context.

**Gates.** `Screenplay.Specs` 1173 passing across net8.0/net9.0/net10.0, unchanged from baseline. `dotnet build -c Release` 41 projects, 0 errors, 1 warning — that warning is pre-existing (`CS0436` on `Program` in `TestApps/Chronicle`), confirmed by stashing and rebuilding the untouched tree. Both end-to-end runs (`TestApps/AspNetCore`, `TestApps/Chronicle`) read back clean, and the generated `.play` files are byte-identical to baseline, so the bump changes no output.

**Two things broke and were fixed:**

1. `ConceptSyntax.Attributes` changed from `IEnumerable<string>` to `IEnumerable<ConceptAttributeSyntax>` in 1.7.0 — the `pii reason` work needed somewhere to put the reason. One call site in `ConceptSyntaxBuilder`.

2. Twelve specs in `for_ScreenplayGenerator/given/a_document_the_language_rejects` failed, and the reason is worth reading. The fixture's "document the language rejects" was a command property named `Description` colliding with the directive — and **1.7.0 fixed exactly that**, so the document is no longer rejected. The stand-in moved to a `query` body, where `description` can only ever be the directive because a query has no properties, keeping the same rejected line and message assertion.

That second one is the argument for staying current: validating against a parser three minors old was not only failing to catch drift, one of the gate's own premises had gone stale underneath it.

**Not done here, deliberately.** The generator still emits none of the constructs 1.7.0 and 1.8.0 added — `type`, `identifier`, `rule`, `performer`, `from $context`. Generating a large production application produces zero of each. Teaching the emitter to use them is separate work; the largest measured wins are `type` (closes 28 unknown-type warnings, the only thing between that application's document and compiling clean) and `rule <Name>` with a `file` reference (addresses roughly 95% of 603 dropped validation rules).

woksin:
Superseded by #2469, now merged. All five commits from this branch are contained in it, and it went further — to Screenplay 2.1.0 rather than 1.8.0, with the authorize requirement tree the new grammar needs.

Closing rather than merging: both would cut two releases for one change.

PR #2469, opened 2026-08-07
Update Cratis.Screenplay to 2.1.0

woksin:
# Summary

Moves the Screenplay language package to 2.0.0 and follows the language where it changed shape, so generated documents keep reading back clean against the compiler that will actually read them.

## Changed

- `Cratis.Screenplay` 1.8.0 to 2.0.0.
- A generated `authorize` block states what it requires on one line, keeping the roles it accepts as alternatives to each other and a named policy as a demand on top of them.
- The value a generated `claim` requirement is matched against is written as a quoted string rather than as a bare word.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -19,7 +19,7 @@
     <PackageVersion Include="Cratis.Chronicle.Testing" Version="16.19.2" />
     <PackageVersion Include="Cratis.Fundamentals" Version="7.17.1" />
     <PackageVersion Include="Cratis.Metrics.Roslyn" Version="7.17.1" />
-    <PackageVersion Include="Cratis.Screenplay" Version="2.0.0" />
+    <PackageVersion Include="Cratis.Screenplay" Version="2.1.0" />
     <!-- MAUI -->
     <PackageVersion Include="Microsoft.Maui.Controls" Version="10.0.90" />
     <PackageVersion Include="Microsoft.Extensions.Logging.Debug" Version="10.0.10" />
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -19,7 +19,7 @@
     <PackageVersion Include="Cratis.Chronicle.Testing" Version="16.16.0" />
     <PackageVersion Include="Cratis.Fundamentals" Version="7.17.1" />
     <PackageVersion Include="Cratis.Metrics.Roslyn" Version="7.17.1" />
-    <PackageVersion Include="Cratis.Screenplay" Version="1.5.2" />
+    <PackageVersion Include="Cratis.Screenplay" Version="1.8.0" />
     <!-- MAUI -->
     <PackageVersion Include="Microsoft.Maui.Controls" Version="10.0.90" />
     <PackageVersion Include="Microsoft.Extensions.Logging.Debug" Version="10.0.10" />
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -19,7 +19,7 @@
     <PackageVersion Include="Cratis.Chronicle.Testing" Version="16.19.2" />
     <PackageVersion Include="Cratis.Fundamentals" Version="7.17.1" />
     <PackageVersion Include="Cratis.Metrics.Roslyn" Version="7.17.1" />
-    <PackageVersion Include="Cratis.Screenplay" Version="1.8.0" />
+    <PackageVersion Include="Cratis.Screenplay" Version="2.0.0" />
     <!-- MAUI -->
     <PackageVersion Include="Microsoft.Maui.Controls" Version="10.0.90" />
     <PackageVersion Include="Microsoft.Extensions.Logging.Debug" Version="10.0.10" />
--- a/Source/DotNET/Screenplay.Specs/IntegrationTesting.cs
+++ b/Source/DotNET/Screenplay.Specs/IntegrationTesting.cs
@@ -67,6 +67,51 @@ public static class IntegrationTesting
             }
         }
 
+        namespace Cratis.Chronicle.Testing.ReadModels
+        {
+            public class ReadModelScenario<TReadModel>
+                where TReadModel : class
+            {
+                public TReadModel? Instance => null;
+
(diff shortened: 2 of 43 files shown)


woksin:
## Reviewer context

### Branch divergence

Two stacks had diverged from the same merge-base with neither containing the other: `upstream/ada-arc` (local only, 22 analyzer/Chronicle commits, Screenplay 1.5.2) and `bump-screenplay-1.8.0` (pushed, 5 Screenplay commits). This branch was cut from `bump-screenplay-1.8.0` and `main` was **merged** into it — no rebase, no force-push, nothing rewritten.

Why that stack: its 5 commits are the Screenplay work this builds on, and it keeps the PR to one concern so it cuts one release. Taking `upstream/ada-arc` instead would have dragged 22 unrelated analyzer/Chronicle commits into a Screenplay bump. That branch is untouched and still holds its work.

**#2462 is superseded by this PR** — 2.0.0 replaces the 1.8.0 bump, and all five of its commits are contained here. It should be closed rather than merged, or the two will cut two releases. I left it open rather than close someone else's PR unprompted.

### Verification

Full solution, both configurations:

```
DEBUG    0 Error(s), 1 Warning(s)   Build succeeded.
RELEASE  0 Error(s), 1 Warning(s)   Build succeeded.
```

The one warning is pre-existing and unrelated — `CS0436` on a `Program` type conflict in `TestApps/Chronicle`'s generated type-discovery provider. Confirmed by stashing this branch's changes and rebuilding that project: identical warning.

All 15 spec projects on `net10.0`:

```
Arc.Core.CodeAnalysis.Specs              Passed  70
Arc.Core.Generators.Specs                Passed  34
Arc.Core.Specs                           Passed  1618
Arc.Specs                                Passed  196
Chronicle.CodeAnalysis.Specs             Passed  41
Chronicle.Specs                          Passed  588
EntityFrameworkCore.PostgreSql.Specs     Passed  34
EntityFrameworkCore.Specs                Passed  300
EntityFrameworkCore.SqlServer.Specs      Passed  34
MongoDB.Specs                            Passed  192
OpenApi.Specs                            Passed  8
Screenplay.Specs                         Passed  1223
Swagger.Specs                            Passed  47
Testing.Specs                            Passed  19
ProxyGenerator.Specs                     Failed  3, Passed 1088
```

`Screenplay.Specs` is green on net8.0/net9.0/net10.0 (1223 each). The 3 `ProxyGenerator.Specs` failures are pre-existing and unrelated — timezone-dependent temporal deserialization (`should_not_shift_the_calendar_day_west_of_utc` and two siblings). Confirmed by stashing this branch's changes and re-running: identical 3 failures.

End-to-end, generating from real applications and reading the document back with the 2.0.0 compiler:

```
AspNetCore.play   64 lines   The generated document reads back clean
Chronicle.play    75 lines   The generated document reads back clean
```

### The one piece that is not mechanical

`AuthorizeSyntax.Policies` became a `PolicyRequirementSyntax` tree. Arc's rule — any one role is enough, a named policy is demanded on top — was previously carried only by the old printer's line layout, never written down. Because `and` binds tighter than `or` in the new grammar, emitting the old flat list would have produced `Librarian or Archivist and SeniorStaff`, which reads back as "any librarian, **or** anyone who is senior staff" and admits callers neither the roles nor the policy admits.

The alternatives are therefore nested under the demand and the printer parenthesizes them. The spec pins this three ways — the tree's shape branch by branch, and the printed surface form:

```
authorize (Archivist or Librarian) and SeniorStaff
```

Neither test application declares `authorize`, so the end-to-end runs above do not cover it; that assertion is the coverage.

### Deliberately out of scope

Screenplay 2.0.0 adds constructs Arc could emit but does not yet — `reads`, `require` inside `validate`, `readmodel`/`reducer`, `for` inside `produces`, and `produces`/`invokes` inside a reactor trigger. Each needs new analysis, emission, and specs, so they are a separate wave. Once they land, these diagnostics can be retired or narrowed:

| Code | Name | Closed by |
|---|---|---|
| `SP0013` | `UnmappableEventSourceIdResult` | `for <expression>` in a `produces` body — directly closes it |
| `SP0020` | `ReducerWithoutCounterpart` | `reducer <Name> => <ReadModel>` with `on <Event>` rules — directly closes it |
| `SP0015` | `UnmappableProjectionConstruct` | partially — a slice holds several projections now, closing the second-projection case |
| `SP0012` | `UnmappableCommandProduction` | partially — `reads` lets a mapping be fed from read-model state |
| `SP0016` | `UnmappableValidationRule` | partially — `require` carries a condition, so a rule held to a `When`/`Unless` can be stated |
| `SP0042` | `ReadModelFeatureWithoutCounterpart` | narrows only — `readmodel` can be declared, but `ReadModelSyntax` carries no tags, which is the case that mattered |

`SP0027` (`UnmappableAggregateStateCondition`) may also relax if `produces when` accepts read-model paths once `reads` is in scope — unverified, do not count it without checking the grammar.

woksin:
Labelled **minor**, deliberately, despite the API change being technically breaking.

`AuthorizeSyntaxBuilder.Build` hands back an `AuthorizeSyntax` that lost `.Policies`, and `SliceContent.IsEmpty` takes a `SliceSyntax` that lost `.Projection` — so by the letter of semver this is major. The reason it is not labelled that way: **the cli is the only real consumer of `Cratis.Arc.Screenplay`, and it is maintained in-house and bumped in lockstep.** A major here would force one on the whole Arc surface — `Cratis.Arc.Core` and the command/query pipeline included — for a change confined to the Screenplay generator, which most Arc applications never invoke.

The residual risk is accepted knowingly: an outside consumer of `Cratis.Arc.Screenplay`'s builders would see a source break under a minor. There is no known such consumer.

Worth revisiting structurally rather than per-release. `Cratis.Arc.Screenplay` re-exports Screenplay's syntax nodes across its package boundary, so every Screenplay major propagates here by construction — the same reason Stage inherited one. Either versioning this package independently of the Arc runtime, or having the builders return Arc's own shapes, would end the recurrence. Neither belongs in an absorption PR.

PR #2470, opened 2026-08-08
Widen ARCCHR0003 and stop discarding Development scope validation

woksin:
# Summary

Arc now preserves the host's dependency-injection validation settings and catches more reactor appends that bypass returned-event routing, without changing explicit non-default event-sequence behavior.

## Changed

- Expand `ARCCHR0003` to report reactors that append to the default event log through an injected `IEventStore`, including indirect member-access paths. (#2480)
- Preserve configured container factories and Development scope validation across the `AddCratisArc` host entry points while disabling only eager validation that Arc cannot support. (#2492)
- Update Chronicle dependencies to 16.20.0. (#2499)

## Fixed

- Allow `Arc.Core` applications to start in Development without discarding the host's other service-provider settings. (#2492)
- Keep `Cratis.Arc.ProxyGenerator.Build` as a package build dependency without flowing its executable assembly into downstream compile and runtime graphs. (#2500)


woksin:
🔴 **Do not merge until Chronicle #3599 has released, and this PR bumps `Cratis.Chronicle` to that release.**

`Directory.Packages.props` still pins `Cratis.Chronicle` 16.16.0. This PR turns Development-time scope validation back on; Chronicle #3599 is what fixes the captive `IEventTypes` in the reactor side-effect handlers. Merging in the wrong order breaks every Arc + Chronicle application in Development — measured in the consuming application as 28 failed reactor partitions, all reporting `Cannot consume scoped service 'Cratis.Chronicle.Events.IEventTypes'`. The source report asked for the two to be sequenced exactly this way.

Reviewer context, kept out of the body so it stays out of the release notes.

**The seam is at three call sites, not one.** Both `UseDefaultServiceProvider` overloads allocate a fresh `ServiceProviderOptions`, so the report's own suggested remedy — "use the `(context, options)` overload and mutate" — does not work; there is no API that hands you the host's current options, so the fix re-derives `ValidateScopes` from `IsDevelopment()`. A consumer wanting a different value calls `UseDefaultServiceProvider` after `AddCratisArc`, which is documented.

**`ValidateOnBuild = false` is still required** and is now pinned by a spec rather than by argument: forcing it true fails `Build()` with 40 inner exceptions from Arc's own contextually-supplied and request-scoped registrations. Arc itself has zero captive dependencies — every registration was resolved from a scope with validation off and on, and the failure sets were identical.

**Arc.Core's fix avoids clobbering.** Applying the container options during `ArcApplicationBuilder` construction rather than in `AddCratisArc` turns an override into a default, so a consumer wiring Autofac or Lamar wins in either call order. An earlier version of this change silently discarded a consumer's `IServiceProviderFactory`; that was caught in review and fixed.

**Two consumer-visible consequences worth a release note.** ARCCHR0003 is a warning, and with warnings-as-errors it now breaks builds carrying the `IEventStore` + `.EventLog` shape — suppression and migration guidance is in `ARCCHR0003.md`. And Development scope validation newly fails any consumer resolving a scoped service from the root provider; two of Arc's own specs do that and pass only because they run in Production.

**One deliberate narrowing:** the analyzer requires the store to resolve to a parameter, field or property, so `var store = eventStore; store.EventLog.Append(...)` is no longer reported. A false positive telling someone to write to the wrong event store is worse than that miss; both are specced and documented.

**Verification.** 38 analyzer specs, each killed by at least one of 14 mutations — no unconditional passes. 11 DI specs across all three hosts covering Development, Production, a custom environment name and both call orderings. Debug and Release build with the one known pre-existing `CS0436` in `TestApps/Chronicle` generated code. 76 + 216 + 1626 specs pass. The analyzer spec harness previously could not tell a broken snippet from a clean one — it now fails on a non-compiling snippet, proven by deliberately misspelling a type in a negative spec.

Issue #2495, opened 2026-08-10
[Resolved] Failed command execution cannot retain a typed success response


woksin:
## Historical problem

At filing, Arc could construct a typed response before execution scopes completed. `TransactionalCommandScope` committed later, and `CommandResult.MergeWith` could retain that response when a commit failure was merged. A caller could therefore receive a success-only value from a command whose transaction did not commit.

## Resolution

**Resolved (2026-08-10):** Fixed in [#2506](https://github.com/Cratis/Arc/pull/2506) and closed as completed. A command that does not succeed no longer carries its response value after execution scopes complete, and the controller path no longer attaches a response alongside exception messages. Four existing specs assert that the response does not survive failure.

## Delivered contract

A success-only typed response cannot escape after final execution-scope completion fails, including transactional commit failure and multi-scope merge ordering.

## What is explicitly not being asked for

Consumer-specific token or HMAC policy is not part of this resolved pipeline contract.

woksin:
Fixed in #2506. A command that did not succeed no longer carries its response value: the pipeline takes it back after every execution scope has completed, and the controller path no longer attaches one alongside exception messages. Four existing specs that asserted the response survived a failure now assert the opposite.

Issue #2476, opened 2026-08-09
Add opt-in authorization revalidation for long-lived observable-query emissions


woksin:
## Problem

Arc runs the query pipeline and all `IQueryFilter` implementations when an observable subscription is created. Later `ISubject<T>` and `IAsyncEnumerable<T>` emissions are written without another authorization decision.

That is compatible with the existing “authorization on every new subscription” contract, but it cannot respond to server-side role, membership, account-state, or resource-access revocation while the connection remains open. A custom query filter cannot close the gap because it has no emission or subscription-invalidation phase, and a browser cannot know that another session changed its access so that it should call `reconnectQueries()`.

## Consumer cost

Consumers requiring prompt revocation must currently wrap every observable, merge authorization-state changes into every stream, build an out-of-band reconnect mechanism, shorten sessions, or avoid observable queries for sensitive data.

## Suggested direction

Add an opt-in, policy-neutral observable-subscription guard owned by Arc. The exact API is open, but it should receive the original query identity/arguments, caller identity, cancellation token, and a fresh scoped provider immediately before an emission is written, and return at least:

- `Allow`
- `Suppress` while retaining the subscription
- `DenyAndTerminate`

It should fail closed on exceptions, work for subject and async-enumerable queries, and behave consistently across direct and multiplexed SSE/WebSocket transports. Arc should remain unaware of application roles or membership stores.

Do not rerun every existing `IQueryFilter` per emission by default: filters may validate, have side effects, or be expensive. A distinct opt-in lifecycle seam preserves their current contract.

## Compatibility

This can be additive and `minor`: preserve current behavior when no guard is registered, add no per-emission work in that case, and make no changes to existing filters or wire contracts.

## Acceptance matrix

1. Current initial-subscription authorization remains unchanged.
2. No guard means exactly current behavior and no new scope/callback.
3. `Allow` works for subject and async-enumerable emissions.
4. `DenyAndTerminate` writes no protected payload, reports unauthorized where supported, and disposes the subscription/enumerator.
5. No emission is written after termination.
6. `Suppress` withholds one emission without advancing the client delta baseline and later `Allow` resumes.
7. A throwing guard fails closed.
8. Multiplexed subscriptions are isolated.
9. Original query name and coerced arguments reach the guard.
10. Scoped dependencies are isolated and disposed with the subscription.
11. Disconnect/cancellation racing with revalidation remains safe.
12. Existing `reconnectQueries()` behavior remains green.
[...]

Issue #2475, opened 2026-08-09
Allow model-bound commands to return exact per-source concurrency scopes


woksin:
## Problem

A model-bound command can read an authoritative state at revision N during its decision, then return one or more `EventForEventSourceId` values. Arc resolves the automatic Chronicle concurrency scope later, in the response-value handler. If another append lands after the decision but before that resolution, Arc can bind the returned events to the newer tail and accept a decision that was made from revision N.

This is expected behavior for Arc's current automatic optimistic strategy, but the command has no supported way to carry the exact revision it actually read into the returned cross-stream event batch.

The lower Chronicle API already supports an `IDictionary<EventSourceId, ConcurrencyScope>` on `AppendMany`; the missing seam is in Arc's model-bound command result path.

## Generic use case

A command reads two authoritative streams, computes a transition, and needs both writes to be conditional on the exact revisions used by that computation. Examples include a transfer plus balance aggregate, a membership transition plus a singleton roster, or any cross-stream invariant whose decision cannot safely be rebound to a later tail.

Today a consumer must encode predecessor revisions into domain events and enforce compare-and-swap through Chronicle value constraints, or leave Arc's command path and append explicitly.

## Suggested additive shape

Add a supported command response/interface that returns:

- the ordered `EventForEventSourceId` values; and
- an optional exact `ConcurrencyScope` per target event source.

Arc should enroll those events and scopes in the existing transactional command unit of work so constraint, concurrency, and commit failures remain ordinary `CommandResult` failures and the entire cross-stream batch stays atomic.

This should be additive and opt-in. Existing event-returning command signatures and automatic tail resolution should remain unchanged.

## Acceptance criteria

- A command can bind each returned event source to a revision captured during its decision.
- Interference after the decision but before Arc handles the return value causes a concurrency failure.
- Cross-stream events and scopes participate in one existing transactional command unit of work.
- A failure leaves no partial append.
- Existing command handlers and wire contracts remain source- and binary-compatible.
- Specs include a timing discriminator where interference before result handling is rejected only when the explicit scope is supplied.

## Non-goals

- Do not change the semantics of Arc's current automatic optimistic strategy.
- Do not add an application-specific cardinality or administrator policy to Arc or Chronicle.
- Do not require callers to bypass the command pipeline or inject `IEventLog` into handlers.

## Resolution

Fixed and released in Cratis Arc 21.9.0.

[...]

PR #2506, opened 2026-08-10
Do not hand back a response a failed command produced, and let a stream be re-authorized

woksin:
# Summary

Two recorded defects in what a command hands back when it fails, and in how long an observable query stays authorized.

## Added

- An opt-in guard consulted before each observable-query emission, so an application can withhold an emission or end a subscription when authorization changes mid-stream. Authorization was previously evaluated once, when the subscription was established, and never again — nothing ended a stream on token expiry, session end or role revocation. With no guard registered nothing is allocated and no dispatch happens, so existing applications are unaffected (#2476)

## Fixed

- A command that did not succeed no longer carries its response value. The value was bound before execution scopes completed, and a commit-time constraint or concurrency failure returned it alongside the failure, serialized into the error body — so a caller could act on a value produced by a command whose transaction rolled back (#2495)
- Model-bound `IAsyncEnumerable` observable queries now work over WebSocket and Server-Sent Events. Both transports previously failed on connect: the observable was cast to an interface it does not implement and the null dereferenced before a single item was streamed

### DIFF
--- a/Documentation/backend/queries/index.md
+++ b/Documentation/backend/queries/index.md
@@ -63,6 +63,7 @@ Every query runs through a pipeline where Arc applies the cross-cutting concerns
 | Authorize by role or policy | [Authorization](../core/authorization.md) |
 | Transform read models before they're served (mask, decrypt, enrich) | [Read Model Interception](./read-model-interception.md) |
 | Stream many live queries over one connection | [Observable Query Hub](./observable-query-demultiplexer.md) |
+| Re-check authorization while a live stream is running | [Observable Query Emission Guards](./observable-query-emission-guards.md) |
 | Debug a live query from the terminal | [Use Observable Queries with cURL](./using-observable-queries-with-curl.md) |
 
 ## The payoff: typed reads in React
--- a/Documentation/backend/queries/observable-query-demultiplexer.md
+++ b/Documentation/backend/queries/observable-query-demultiplexer.md
@@ -166,6 +166,8 @@ Authorization is enforced for every subscription through the standard query pipe
 - If the query allows anonymous access (`[AllowAnonymous]`), the subscription is accepted regardless of authentication state.
 - Authorization is re-evaluated on every new subscription, not cached for the lifetime of the connection.
 
+That verdict gates *obtaining* the stream. A subscription can then stay open indefinitely, and nothing ends it when a token expires or a role is revoked. To re-check the verdict while a stream is running, implement an [emission guard](./observable-query-emission-guards.md).
+
 ## Keep-alive
 
 Both WebSocket and SSE transports send automatic keep-alive messages to prevent idle connections from being closed by proxies or firewalls.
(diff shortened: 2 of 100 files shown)


woksin:
Reviewer context, kept out of the description because it is not release notes.

**Verification.** `dotnet build` Debug and Release both 0 warnings / 0 errors, no generated-proxy churn. `dotnet test`: Arc.Core.Specs 1,800 passed / 0 failed (baseline 1,763), Arc.Specs 224 / 0, every other assembly green. TypeScript: `tsc -b` clean, lint clean, vitest 765 passed (baseline 761).

The 3 `ProxyGenerator.Scenarios.for_ProxyGeneration.when_deserializing_temporal_properties_through_the_generated_proxy` failures are pre-existing ClearScript/V8 failures unrelated to this change. I confirmed that directly: checked out `4cb52036` untouched and ran that assembly in isolation — identical 1,088 passed / 3 failed.

Every item was verified by mutation — reverting the fix must turn a named spec red — and restored by hand before the next step.

**Semver ruling, stated rather than left silent.** Five public constructors gained parameters (`ClientObservable`, `ClientObservableSSE`, `ClientEnumerableObservable`, `ClientEnumerableObservableSSE`, `ObservableQueryDemultiplexer`). On a strict reading that is `major`. I labelled `minor` on this repo's own precedent: `ClientObservable<T>`'s arity has changed three times already (`030b6bfd`, `b7444058`, `82d76ad6`), these types are only ever constructed through `ActivatorUtilities`, and the public factory surface in `ObservableQueryExtensions` is unchanged. Happy to relabel if the maintainers read it differently.

**Four existing specs assert the opposite of what they used to, deliberately.** They set up a failing value handler and asserted the response survived it — i.e. they pinned the defect. Each keeps its `should_set_response_on_command_context` assertion, so which tuple element becomes the response is still pinned.

**Independent code and security review found ten findings on this branch.** The blocking one and the security hold-merge item are fixed in the last commit: a denial on a replayed subject's first emission leaked the subscription scope, semaphore and observer on SSE; and a denial on the direct SSE transport reconnect-looped, re-running the whole query every few seconds.

**The async-enumerable fix was found by writing specs, not by reading.** The four `ActivatorUtilities.CreateInstance` sites in `ObservableQueryHandler` had zero coverage. Covering them exposed that both async-enumerable sites cast to `IClientObservable`, an interface `IClientEnumerableObservable` has no relationship to — so the cast always yielded null and the next line dereferenced it. Broken on shipped main over both transports.

**Deliberately not done, so it is a decision rather than an omission.**
- The security reviewer proposed mapping "the guard could not answer" to `Suppress` rather than `DenyAndTerminate`, so a transient session-store fault would not permanently end every live query. The reasoning is sound but it changes the security contract of a seam shipping for the first time. The blast radius is documented instead: a guard that throws denies, the aggregator is a process-wide singleton, so one unconstructable guard type denies every observable query in the application until it is fixed.
- The multiplexed async-enumerable path still runs no `InterceptEmission`, so PII release does not apply there. Pre-existing and unchanged by this train — worth its own issue.
- IMP-36 (#2475) is intentionally not in this train. It needs the Chronicle 16.22.0 bump in #2504 and an owner decision on whether a target-less concurrency scope can be enrolled in the command transaction at all.

Issue #2493, opened 2026-08-10
CommandForm applies explicit undefined values over resolved data


woksin:
## Describe the bug

This issue promotes a previously recorded upstream item from a production Cratis consumer. The complete evidence and requested fix follow.

**Status:** Fixed in [#2510](https://github.com/Cratis/Arc/pull/2510) and shipped in [v21.5.0](https://github.com/Cratis/Arc/releases/tag/v21.5.0) · closed 2026-08-11
**Product:** Arc — `@cratis/arc.react` `CommandForm`
**Type:** Correctness / destructive pre-population

## Symptom and Ada cost

`CommandForm` distinguishes presence from neither value nor intent: if `initialValues` or `currentValues` contains
a key whose value is `undefined`, the form applies it over the command's already-resolved value instead of treating
it as absent. Ada shipped this through an address flow: an async registry value was resolved correctly and then
wiped by a present-but-undefined overlay. Ada now carries two ESLint rules (`no-undefined-over-resolved-value` and
`spread-before-defaults`) to keep the trap unreachable.

## Requested fix

Define and enforce one safe overlay contract. Prefer skipping `undefined` entries while preserving explicit
nullable/clear values, or expose a typed sentinel for intentional clearing. Add cases for direct object literals,
prop-spread wrappers, async `currentValues`, and already-resolved command values.

## Resolution

Fixed in [#2510](https://github.com/Cratis/Arc/pull/2510) and shipped in v21.5.0. `CommandForm` now ignores an `initialValues` key whose value is `undefined` when it would otherwise overwrite a value already resolved through `currentValues`; seed entries contribute only keys carrying a value. This issue does not establish a broader intentional-clear or `currentValues`-with-`undefined` contract.

woksin:
Fixed in #2510, shipped in [v21.5.0](https://github.com/Cratis/Arc/releases/tag/v21.5.0) (merge `123b81f1`).

`CommandForm` no longer lets an `initialValues` key holding `undefined` overwrite a value that `currentValues` had already resolved. The seed layer now contributes only keys that actually carry a value, so a resolved field survives a later render in which the seed object still lists the key as `undefined`.

Issue #2473, opened 2026-08-09
Return a terminal failed result when a useQuery transport request fails


woksin:
## Problem

In Arc v21.0.4, `QueryFor.perform()` awaits `executeQueryHttpRequest()` outside its response-parsing `try/catch`. A rejected `fetch()` therefore rejects the performer promise. `useQuery` catches that rejection with a bare catch, tells `QueryScope` execution completed, but never replaces `QueryResultWithState.initial`, whose `isPerforming` remains `true`.

A dead network consequently renders as a query loading forever while the surrounding scope says it completed.

This report is specifically about `useQuery`; it does not claim the same mechanism for `useObservableQuery`.

## Evidence

- `useQuery.ts`: initial result, performer call, bare catch, completed notification.
- `QueryResultWithState.ts`: initial state has `isPerforming: true`.
- `QueryFor.ts`: transport await is outside the catch; only response parsing becomes `noSuccess`.
- Existing `with_fetch_error` catches the rejection in the spec itself and manufactures `QueryResult.noSuccess`, so it does not prove runtime handling.

## Acceptance criteria

- A non-abort transport rejection produces a terminal unsuccessful result with `isPerforming == false` in the hook and query scope.
- The lower-level fetch-error spec passes without an outer test catch or synthetic result.
- An aborted superseded request cannot overwrite the newer request.
- Re-execution can replace the failure with success.
- Failure caching behavior is explicit and covered.

No public signature needs to change; this is compatible corrective behavior.

woksin:
Release-train note: please group this compatible JavaScript fix into the current/next Arc minor train rather than opening a one-fix release PR. No public signature change is needed.

woksin:
Fixed in #2510, shipped in [v21.5.0](https://github.com/Cratis/Arc/releases/tag/v21.5.0) (merge `123b81f1`).

A query that fails in transport now reaches a terminal, unsuccessful result instead of reporting that it is still performing forever:

- `QueryFor.perform` resolves with an unsuccessful result when the request never reaches the server.
- `useQuery` settles as unsuccessful and carries the exception message.
- `useSuspenseQuery` surfaces the failure through its error boundary as `QueryFailed`, where it previously left the boundary pending indefinitely.
- Only a successful result is cached, so a failed attempt no longer seeds a later mount of the same query.
- An aborted, superseded request can no longer settle over the newer request that replaced it.

[...]

Issue #2485, opened 2026-08-10
Add regression specs for CommandForm silent-validation ordering guard


woksin:
## Summary

`CommandForm` already rejects a late silent-validation result that a newer validation has overtaken. This issue identified that the ordering guard lacked regression coverage; it did **not** require a production behavior change.

Closed by [#2510](https://github.com/Cratis/Arc/pull/2510), shipped in [Arc v21.5.0](https://github.com/Cratis/Arc/releases/tag/v21.5.0). The change adds specifications designed to fail if the existing ordering guard is removed or bypassed.

## Why the regression coverage matters

A stale validation result must not decide whether a `CommandForm` is valid after the user has changed its values. If that invariant regresses, a submit action can remain disabled even though the current values are valid, without a useful error for the consumer.

The original investigation observed behavior consistent with this failure and traced the silent-validation path. Upstream review established that the production guard was already present and correct; the missing artifact was a falsifiable test of that guard.

## Requested seam at filing

Provide regression specifications for the silent-validation ordering invariant: a result that has been overtaken by a later validation must not update form validity.

## Acceptance

The test suite fails when the ordering guard is removed or bypassed, and passes with the intended guard in place.

## What is explicitly not being asked for

- Not a request to change `CommandForm` validation behavior or policy.
- Not a request to remove client-side short-circuiting.
- Not a request for consumer-specific validation or UI behavior.

woksin:
Closed by #2510, shipped in [v21.5.0](https://github.com/Cratis/Arc/releases/tag/v21.5.0) (merge `123b81f1`).

**No production change was needed** — the ordering guard against a late-arriving stale validation response already existed and behaved correctly. What was missing was any spec that could fail if it stopped working, so the guard was unfalsifiable and one refactor away from silently disappearing.

The commit `0025cfd3` adds specs that pin it against its own blind spots: each assertion was checked to die to a single named production line, rather than merely passing alongside a correct implementation. Because this arm of the change is spec-only, it carries no user-facing release-note bullet — which is why this issue number does not appear in the published notes for v21.5.0.

PR #2510, opened 2026-08-11
Settle failed queries and stop a seed undefined wiping resolved data

woksin:
## Changed

- `QueryFor.perform` resolves with an unsuccessful result when the request never reaches the server, instead of rejecting (#2473)
- A query that fails in transport now surfaces through `useSuspenseQuery`'s error boundary as `QueryFailed`, where it previously left the boundary pending forever (#2473)
- Only a successful query result is cached, so a failed attempt no longer seeds a later mount of the same query (#2473)

## Fixed

- `useQuery` settles as unsuccessful when a query fails in transport, carrying the exception message, instead of reporting that it is still performing forever (#2473)
- An aborted, superseded query request can no longer settle over the newer request that replaced it (#2473)
- `CommandForm` no longer lets an `initialValues` key holding `undefined` overwrite a value that `currentValues` resolved (#2493)


woksin:
Reviewer context — none of this belongs in the release notes.

### Commits

| SHA | Item | Kind |
|---|---|---|
| `a159073a` | ARC-29 / #2473 | fix + specs |
| `9306bfcc` | ARC-34 / #2493 | fix + specs |
| `0025cfd3` | ARC-28 / #2485 | **spec-only** — the production guard shipped in 21.0.0 (`c77d43d0`) |

Grouped into one train per the request on #2473 to avoid a one-fix release PR.

### Semver reasoning — `minor`, and why not `patch`

No public signature changes and nothing new is exported: `isAbortError` is an internal module, and `@cratis/arc`'s `exports` map has only barrel subpaths (`.`, `./commands`, `./queries`, …), so it cannot be deep-imported. `withoutUndefinedValues` is internal to `Arc.React`.

`patch` was the obvious reading, but three behaviors change in ways a consumer can observe, so `minor` matches how #2473 is already labelled. **The one worth a maintainer's eye:** `QueryFor.perform()` no longer rejects on a transport failure. A consumer doing `try { await query.perform() } catch { … }` silently stops entering its catch and must read `isSuccess` instead. Downgrade or escalate if you read that differently.

### Verification

Gates re-run by the orchestrator, not just reported: Arc **300 files / 769 tests**, Arc.React **178 / 467** (baseline 166 / 431), Arc.React.MVVM **9 / 14**. Build + `lint:ci` clean in all three. Working tree clean, no `dist/` staged.

Two mutations run independently of the implementers:

- `QueryFor.ts:150` `isSuccess: false` → `true`, **with a rebuild** → exactly one assertion red (`should not be successful`), other four green.
- `CommandForm.tsx:226` `return false` → `return true` — the boolean only, guard still skips the write → **5 assertions red across 5 files**, two of them in the production message path. Before `0025cfd3` that mutation was invisible to the entire suite.

A methodology note for anyone re-running this: `Arc.React` resolves `@cratis/arc` through the `exports` map to `dist/`, so mutating `Arc` source without `yarn build` in `Arc` tests the stale build and yields a **false green**.

### Cache decision (#2473, criterion 5)

Failures are deliberately not cached. `QueryInstanceCache` exists to hand the last known good payload to future subscribers; caching a failure would poison every mount for the 30s retention window. Not caching means a remount shows stale-but-good data and re-fetches — stale-while-revalidate. This became load-bearing once the fix made transport failures *resolve* through the normal path, which previously reached `setLastResult`.

### On #2493 — what was deliberately not done

The report's preferred remedy was to skip `undefined` everywhere. That would revert a deliberate, specced, released contract: `1906ab23` ("Allow currentValues to clear command properties") replaced `!== undefined` with `hasOwnProperty` in both `CommandForm.tsx` and `useCommand.ts`, pinned by `when_currentValues_changes.ts:131` and `for_useCommand/when_setting_values.ts:66`.

So only the **seed** layer changed; the reactive layer keeps presence semantics. A uniform contract across both layers is a **major** and is a maintainer decision, not one taken here. A fence spec keeps the `currentValues` clear-contract alive against a later over-broad fix.

### A seventeenth vacuity species, found while writing these specs

**Absorbing-default vacuity** — asserting a value that equals the observable's *not-yet-determined* state. `isValid` is `false` before the first silent validation answers *and* after a failed one, so `await waitFor(() => expect(isValid).toBe(false))` is satisfied at t=0 whether or not anything ever validated: a spec that looks asynchronous, passes in ~0 ms, and proves nothing.

It was caught by a detector worth keeping: **a surviving assertion whose sibling died to the same mutation**. Those assertions now read `validationMessages`, which only a *completed* run populates. Every negative message assertion is also paired with a positive control proving the harness renders that message when nothing overtakes the run — otherwise "no stale message" is satisfied by a form that renders no messages at all.

### Not pinned — stated plainly

- `CommandForm.tsx:227` (the high-water-mark advance) has no assertion that dies to it *alone*; everything it kills is also killed by `:226` or by the inverted-comparison mutation. Inherent — `:227` exists to make later results stale, which is what `:226` reads. `:226` **is** now independently falsifiable, so the pair is broken up.
- `revertChanges()` cannot discriminate the #2493 fix: it assigns `undefined` before and after. Change tracking only discriminates in the static shape, which is where it is pinned.
- W2×W2 (two keystroke validations racing) is **not covered and not claimed** — `validateClientSide()` is synchronous, so issue order always equals resolve order and such a spec would be vacuous by construction.
- The two abort specs are regression guards, not red-before-green: pre-fix, `perform()` rejected on everything and `useQuery` swallowed everything, so abort behavior was accidentally correct.

### Out of scope, found on the way

The 13 pre-existing `for_useQuery` specs stub `fetch` with a body carrying no `paging`, so `new QueryResult(...)` throws and silently falls into `noSuccess`. **Those specs have never once seen a successful query result.** Not touched here — it wants its own issue.

`QueryHttpRequest.ts` shows as `Bin` in diffs because it contains a literal NUL byte used as a cache-key separator in `backendKey`. Pre-existing at `HEAD`, unrelated to this PR, but it makes that file's diffs unreadable and would be better as a `\0` escape.

woksin:
**Review response — `7db41d51`.** Both reviews resolved; one of my own claims was wrong and is corrected below.

### MUST-FIX 1 — the co-subscriber left performing forever: fixed

The abort branch now settles functionally instead of bare-returning. Pinned by a new two-component spec sharing one `QueryInstanceCache`; the first request's stub honours its `AbortSignal`, so the abort is genuinely produced by the second component superseding the first through the shared instance — nothing is simulated, and if instance sharing ever broke the spec fails rather than passing vacuously.

**Correction to my own prescription.** I wrote that the functional form means "if the newer request already settled, this takes *that* result". That is only true when a hook supersedes **itself**. `useQuery` reads the cache **only** in its `useState` initializer and never registers a listener, so a superseded co-subscriber sees its own state, not the sibling's.

Net effect: the co-subscriber settles `isPerforming: false`, `isSuccess: true`, `data = defaultValue`. **The forever-performing spinner is gone**, but it renders its empty state rather than adopting the sibling's payload. Making it adopt that payload means subscribing the hook to cache listeners — a real design change that was deliberately not smuggled in here. The reviewer's second probe assertion ("the first component should have the data") therefore cannot pass with this fix, and was not written. **Worth a follow-up issue.**

### SHOULD-FIX 2, 3, 4, 5, 6 — all addressed

- Nullish rejection: `(error ?? {})` in both places. It threw while destructuring, which made the `String(error)` fallback written for that exact case unreachable and broke the resolve-instead-of-reject contract.
- `hasData`: the literal is now returned via `Object.setPrototypeOf(failure, QueryResult.prototype)`. Chosen over `new QueryResult(...)` deliberately — the constructor runs `data` through `deserializeQueryModel*`, which would change `data` from `this.defaultValue` (same reference) to a deserialized copy, and to `null` for a non-enumerable query with a nullish default. Field values are byte-for-byte unchanged.
- `useSuspenseQuery`: the release-note bullet now has a spec.
- The data-wipe axis is pinned. Behaviour **unchanged** — carry-previous-data is a UX decision and did not ride in silently.
- The query cache key separator is now written as an escape rather than a literal NUL.

**Honest limit on the NUL fix:** it does **not** un-hide this PR's own diff. Git decides binary-ness per blob and needs both sides textual; `HEAD`'s blob still holds the NUL, so the file still renders `Bin` here. Every diff from the landing commit onward is reviewable. Runtime value verified unchanged: the composite key is still 14 characters with the separator at code point 0.

### Verification

Gates re-run by the orchestrator: Arc **301/773**, Arc.React **181/476**, MVVM **9/14**, build + `lint:ci` clean in all three, tree clean. Baselines were 300/769 and 178/467 — exactly the specs added.

Independent mutation: reverting the abort branch to the bare `return` reddens **exactly** `should stop the superseded component performing` and nothing else.

### A methodological finding from the implementer, worth recording

The first attempt to pin the data-wipe axis mutated `useQuery`'s `failedResult(queryInstance.defaultValue, …)` — and the spec **stayed green**. After this train a transport failure never reaches `useQuery`'s catch at all: `QueryFor.perform()` resolves with the failure result, so `data` is blanked in `QueryFor`. Trusting the obvious mutation site would have recorded a false negative.

Related consequence: `useQuery`'s `failedResult` is now reachable **only** when a performer rejects with something that is not an `AbortError`, which no `QueryFor` subclass does any more. That is why `FakeNullishRejectingQuery` exists — without it the whole helper is unreachable from every shipped code path.

### Still open for a maintainer — the label

`minor` vs `major` on `QueryFor.perform()` no longer rejecting. Unchanged by this commit and still the one thing gating merge.

### Out of scope, confirmed present

The same prototype-dropping spread of `QueryResult.noSuccess` pre-exists at `QueryFor.ts:85` (the invalid-arguments and JSON-parse-failure paths) and in `ObservableQueryFor.ts`. Untouched here.

Issue #2501, opened 2026-08-10
Preserve ClientPrincipal.IdentityProvider when reconstructing the principal


woksin:
## Describe the bug

AuthProxy serializes `ClientPrincipal.IdentityProvider`, but Arc's `MicrosoftIdentityPlatformAuthenticationHandler` reconstructs the ASP.NET principal from the claims, user id, name, and roles and drops that field. The resulting authentication type is Arc's generic handler name rather than the forwarded scheme metadata.

The released AuthProxy value is display/scheme-derived, so it is not a durable provider key and must not be repurposed as one. The defect is wire fidelity: a public field the producer emits has no consumer-visible representation after the normal Arc authentication boundary.

## Expected behavior

Preserve the forwarded identity-provider/scheme metadata through a documented claim or feature while keeping it distinct from AuthProxy's canonical provider-key, issuer, and subject contract.

- Remove or reject colliding case-varied claims before writing the trusted value.
- Do not infer membership, issuer, or provider identity from the display-derived field.
- Keep the existing canonical claims introduced by AuthProxy 2.13 independent.

## Acceptance criteria

Feed two serialized `ClientPrincipal` values that differ only in `IdentityProvider` through the real Arc handler. The reconstructed principals expose the two exact scheme values. Duplicate or case-varied attacker-authored claims cannot override the trusted transport value.

## Evidence and impact

The omission is source-confirmed at Arc 21.0.4 and AuthProxy 2.11.4. AuthProxy 2.13 now carries a separate canonical identity tuple, so this is no longer an Ada launch dependency; it remains a compatibility and diagnostics defect in Arc's consumption of the released wire type.

Originating consumer record: Ada ARC-31.

woksin:
Fixed in #2511 (`310ccc2cfcf5e662e0ac1220ada5e92e8e2dde5e`).

Both handlers now carry `ClientPrincipal.IdentityProvider` as a reserved claim, `urn:cratis:arc:identity:provider`. It reaches `IdentityProviderContext.Claims` through the projection that already forwards every principal claim, so it is available to `IProvideIdentityDetails.Provide` without new plumbing or a new DTO.

It is carried as a **claim** rather than as `AuthenticationType` deliberately — stamping the scheme name would break anything matching on it, which would have made this a major.

[...]

PR #2511, opened 2026-08-11
Carry the identity provider the ingress authenticated a caller with

woksin:
## Added

- The identity provider the ingress authenticated a caller with is carried on the authenticated principal as a reserved claim, available to the identity details an application provides (#2501)

### DIFF
--- a/Documentation/backend/asp-net-core/microsoft-identity.md
+++ b/Documentation/backend/asp-net-core/microsoft-identity.md
@@ -45,6 +45,45 @@ app.UseAuthentication();
 app.UseAuthorization();
 ```
 
+## Knowing which identity provider signed the caller in
+
+The `x-ms-client-principal` payload carries an `identityProvider` field describing the provider the ingress
+authenticated the caller with. Without it, everything downstream sees is a set of claims that look the same whether
+they came from Entra ID or GitHub — so the application has no way to tell one federation apart from another.
+
+Arc keeps that value on the reconstructed principal as a claim, so both normal request authorization and the
+`/.cratis/me` identity resolution can read it:
+
+```csharp
+using Cratis.Arc.Identity;
+using Microsoft.AspNetCore.Http;
+
+public class IdentityProviderReader(IHttpContextAccessor httpContextAccessor)
+{
+    public string? Current =>
+        httpContextAccessor.HttpContext?.User.FindFirst(MicrosoftIdentityPlatformClaims.IdentityProvider)?.Value;
+}
+```
+
+| Aspect | Detail |
+| ------ | ------ |
+| Claim type | `urn:cratis:arc:identity:provider` — use the `MicrosoftIdentityPlatformClaims.IdentityProvider` constant |
+| Value | The exact `identityProvider` value the ingress forwarded, for example `aad` or `github` |
+| When absent | The forwarded principal carried no identity provider |
+
+The claim type is **reserved for Arc**. The `x-ms-client-principal` header is base64, not a signature, so any caller
+that can reach the application can put whatever it likes in the serialized `claims` array — including a claim of this
+very type. Arc therefore removes every claim of the reserved type from the deserialized payload, ignoring casing,
+*before* writing its own value. A caller cannot smuggle in an identity provider of its own choosing, and when the
+forwarded principal carries no identity provider the claim is simply absent rather than attacker-supplied.
+
+> [!IMPORTANT]
+> This value is authentication-scheme metadata and is typically derived from a provider display name. It is useful for
+> diagnostics, for telling federations apart, and for provider-aware behavior — but it is not a stable provider
+> registration key, and renaming a provider can change it. Do not use it as a durable identity, membership or storage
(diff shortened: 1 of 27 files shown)


woksin:
Reviewer context — resolves #2501. Commit `9f3507d4`.

### Why a claim, and not `AuthenticationType`

The obvious fix is to stamp `AuthenticationType` from `ClientPrincipal.IdentityProvider`. That was rejected: anything matching on the scheme name breaks, which would make this `major`. Carrying it as a claim is additive, and `IdentityProvider.cs:191` already projects **every** principal claim into `IdentityProviderContext.Claims`, which reaches `IProvideIdentityDetails.Provide` and `/.cratis/me` — so the "available to both" requirement needed no new plumbing and no new DTO.

New public surface is exactly one constant, hence `minor`.

### The namespace choice is load-bearing

`urn:cratis:arc:identity:provider` sits deliberately **outside** `urn:cratis:identity:`, which AuthProxy reserves for its own canonical claims (`CanonicalIdentityClaims.ProviderKey/Issuer/Subject`). AuthProxy strips that whole prefix from a submitted principal before cookie storage — there is a spec for it — so keeping Arc's claim out of that namespace means Arc can never overwrite the ingress's **stable provider key**. The value Arc carries is descriptive (typically a provider display name), not durable, and the docs say so explicitly: do not use it as an identity or membership key.

### The forgery surface, and why the strip is case-insensitive

A reserved claim type is forgeable — a submitted principal can carry one of its own. Both handlers strip it before writing the trusted value.

The strip is `OrdinalIgnoreCase` **by design**: `ClaimsIdentity.FindAll` matches case-insensitively, so an ordinal-only strip would leave `URN:CRATIS:ARC:Identity:Provider` findable. That is the actual bypass, and it has its own spec arm. With no forwarded provider, **no claim is written at all** rather than an attacker's surviving.

The legacy handler (`Source/DotNET/Arc/Identity/MicrosoftIDentityPlatformAuthHandler.cs`) dropped the field identically and is fixed the same way.

### Verification

Re-run by the orchestrator, not taken on trust: **Release `--no-incremental`, 0 warnings / 0 errors**; `Arc.Core.Specs` **1835**, `Arc.Specs` **252**, all passing. 28 new facts in each project.

Independent mutation — dropped the reserved type from the core handler's `RemoveAll`:

```
Failed …carries_a_colliding_identity_provider_claim.should_expose_the_identity_provider_the_ingress_supplied
Failed …carries_a_colliding_identity_provider_claim.should_expose_exactly_one_identity_provider
Failed …carries_a_colliding_identity_provider_claim.should_strip_the_forged_identity_provider
Failed …colliding_claim_only_differs_in_casing.should_expose_the_identity_provider_the_ingress_supplied
Failed …colliding_claim_only_differs_in_casing.should_expose_exactly_one_identity_provider
Failed …colliding_claim_only_differs_in_casing.should_strip_the_case_varied_forgery
Failed …no_identity_provider_but_a_colliding_claim.should_not_expose_an_identity_provider
Failed! - Failed: 7, Passed: 21, Total: 28
```

Exactly the forgery arms; the other 21 unaffected. Each of the four guards — the strip, its case-insensitivity, the write, and the empty-guard — was mutated independently in **both** handlers and dies alone.

### A twentieth vacuity species, found and mitigated

**A negative assertion that a *more broken* unit also satisfies.** `should_not_expose_an_identity_provider` is green when the strip works — and equally green when the handler drops **every** claim. So each arm carries a sibling baseline pin (`should_keep_the_unrelated_claim`), and a `RemoveAll(_ => true)` probe confirms those pins are load-bearing: under it, both absent-provider arms go red, so their negatives are never alone.

Distinct from absorbing-default vacuity (#17), where the asserted value equals the observable's *not-yet-determined* state. Here the value is produced — by over-deletion.

Also applied species #19 (the unpinned enabling condition): the spec's precondition is that `identityProvider` genuinely deserializes off the wire, so the principal header is built from a **literal Azure/AuthProxy wire payload** verified against `AuthProxy/Source/AuthProxy/Identity/ClientPrincipal.cs`, not round-tripped through Arc's own serializer. A naming-policy regression reddens rather than passing self-consistently.

### Deliberate deviation

Two `Documentation/**` files are included. `framework.md`'s quality gate requires docs for public-facing API changes, and a reserved claim type is worthless undocumented — the security contract ("Arc strips it; do not use it as a durable key") only exists if it is written down. Nothing else depends on them.

### Not verified

- Docs not rendered — `npm run check` lives in the sibling `Documentation` repo, outside this repo's write scope. Content self-checked against the formatting rules and every API in the snippet verified against source.
- No live AuthProxy↔Arc pair was exercised. AuthProxy's real wire shape and reserved namespace were read from source and encoded into the specs, but nothing ran a real proxy against a real Arc host.
- The `/.cratis/me` seam is asserted by construction, not by a spec crossing it — that is a different unit, and a spec there would be a fence in a neighbour's yard.

Related: AuthProxy #96 ("Accept Microsoft multi-tenant sign-ins and attribute providers by scheme") is the producing side of this — it is what makes the forwarded value meaningful.

Issue #2471, opened 2026-08-09
Subdomain tenant resolver treats apex domains and IP literals as tenant prefixes


woksin:
## Describe the bug

`SubdomainTenantIdResolver` treats every dotted host as having a tenant subdomain. This contradicts the resolver's documented and PR-defined fallback behavior for an apex application host.

The current implementation splits the host and returns the first label whenever there is more than one segment. Consequently:

- `myapp.com` resolves tenant `myapp` instead of the configured header fallback;
- `www.myapp.com` resolves tenant `www`;
- `127.0.0.1` resolves tenant `127`.

The resolver has no configured base-domain/public-suffix knowledge with which to distinguish `acme.myapp.com` from `myapp.com`, so the advertised contract cannot be implemented reliably from segment count alone.

## Steps to reproduce

1. Configure `options.UseSubdomainTenancy("X-Tenant-Id")`.
2. Supply host `myapp.com` and header `X-Tenant-Id: hive-consulting`.
3. Resolve the tenant with `SubdomainTenantIdResolver`.

## Expected behavior

The behavior promised by the documentation and introducing PR is honored: an apex application host falls back to the configured header, while a genuine tenant host such as `acme.myapp.com` resolves `acme`.

The configuration should explicitly state the application/base domain (or use another unambiguous contract) so the resolver can distinguish an apex host from a tenant prefix. IP literals, localhost, malformed/empty hosts, multiple header values, `www`, mixed casing, ports, and internationalized hosts need explicit fail/fallback behavior and specifications.

## Actual behavior

Any dotted host returns its first segment as the tenant. Existing specs cover `acme.localhost` and single-label `localhost`, but not the documented `myapp.com` apex case.

## Environment

- Arc: v21.0.4 (`dc7e9e362969fd2139b119e67e9bff3167aeb524`)
- .NET: 10

## Additional context

- Current implementation: `Source/DotNET/Arc.Core/Tenancy/SubdomainTenantIdResolver.cs`
- Contradicting documentation: `Documentation/backend/tenancy/resolvers.md`
- Introduced by commit `115d60d13f4c2852f389165cd991ad0f624ba0d1` / PR #2198, whose stated behavior is “header fallback when no subdomain is present.”

The fallback itself was deliberate; the segment-count implementation cannot identify the condition it promises. This is therefore a runtime/design defect rather than a request to change an intentional apex-domain policy.

woksin:
[...]

Issue #2494, opened 2026-08-10
generated command setters are non-nullable for optional C# properties


woksin:
## Describe the bug

This issue promotes a previously recorded upstream item from a production Cratis consumer. The complete evidence and requested fix follow.

**Status:** Fixed in [#2519](https://github.com/Cratis/Arc/pull/2519) and shipped in [v21.6.1](https://github.com/Cratis/Arc/releases/tag/v21.6.1); closed as completed on 2026-08-11. Consumer applications should regenerate proxies and re-run strict TypeScript checks.
**Product:** Arc proxy generator — TypeScript commands
**Type:** Type fidelity / generated API

## Symptom and Ada cost

For an optional C# property (`T?`), the generated interface correctly declares `x?: T`, but the generated command
class getter/setter remains `T`. Clearing the value in `onBeforeExecute` therefore fails TypeScript compilation
even though the descriptor is optional and omits `undefined` from the request. Ada has to cast the generated class
through its optional interface, or introduce a parallel `CommandLike` interface, in two established idioms.

## Requested fix

Make the generated class property type agree with its optional interface and wire descriptor (`T | undefined`),
with generator specs covering read, assignment-to-undefined and omission from the serialized request.

## Honest limits

The consumer compile failure and wire omission are observed. The exact generator emission site has not been
re-derived in this filing pass.

woksin:
Fixed in #2519, shipped in [v21.6.1](https://github.com/Cratis/Arc/releases/tag/v21.6.1) (merge `5448a4e9`).

`Templates/Command.hbs` already knew which properties were nullable — it used the same `IsNullable` on the wire descriptor — but the backing field and accessors ignored it, so a nullable command property generated as a non-nullable one. The proxy's type asserted a value is always present when it is not. The accessors now carry `| undefined`, matching the declaration.

Shipped as a **patch**: the generated type was wrong and correcting it restores what should have been there from the start. Nothing observable moves — the wire format is byte-identical, `PropertyDescriptor(name, type, isNullable)` was already correct and is untouched, and no public C# surface changed.

One consumer-visible effect, worth stating: an application using `strictNullChecks` that reads an optional command property into a non-nullable position will get TS2322/TS2345 when it regenerates. That error is the fix working — it points at a latent bug the incorrect type was hiding.

[...]

PR #2517, opened 2026-08-11
Resolve subdomain tenancy against a required base domain

woksin:
## Added

- `TenancyOptions.BaseDomain`, configured through `UseSubdomainTenancy(baseDomain, fallbackHeaderName)` (#2471)
- `TenancyOptionsValidator`, which validates tenancy configuration while the host starts (#2471)

## Changed

- Subdomain tenancy resolves a tenant only for a host that sits below the configured base domain, and falls back to the tenant header for anything else — an apex domain, a deeper host, or an address (#2471)
- A host is matched case-insensitively and in punycode form, so one domain is always one tenant regardless of how it was written (#2471)
- A resolved tenant must be a valid DNS label, so a host cannot put a wildcard, user information, a control character or an unbounded string into a tenant identifier (#2471)

## Fixed

- An apex domain no longer resolves to a tenant named after the site (#2471)
- An IPv4 or IPv6 literal is no longer read as a tenant, in any spelling (#2471)
- `IHttpRequestContext.Host` no longer truncates an IPv6 literal to its opening bracket (#2471)

## Removed

- The `UseSubdomainTenancy(fallbackHeaderName)` overload — subdomain tenancy now requires a base domain, and the header name is the second argument (#2471)

## Security

- Subdomain tenancy configured without a valid base domain now fails at startup, instead of resolving no tenant from any host and falling back to a client-supplied header that selects the namespace and database (#2471)


woksin:
Reviewer context, kept out of the body. Two items, bundled deliberately so consumers absorb one breaking upgrade rather than two.

## Semver reasoning, per item

**ARC-33 (#2471) on its own would be arguable.** It is source-compatible — `TenancyOptions.BaseDomain` and one overload are additive, nothing is removed or renamed, and every existing call site compiles unchanged. It is behavior-breaking only for an application that registers `UseSubdomainTenancy()` and relies on `acme.<anything>` resolving `acme`; after this it gets the header fallback until it sets a base domain. `UseSubdomainTenancy` has **zero call sites** across Arc, AuthProxy, Chronicle, Components and the consuming application that reported this, so the measured blast radius is nil. Downstream consumers outside those repositories are unmeasured.

**ARC-35 (#2494) is what makes this `major`.** The generated proxy shape is public API and every application regenerates it on a Release build. An optional command property's getter widens from `T` to `T | undefined`, so a consumer reading it into a non-nullable position gets TS2322/TS2345 with no code change of their own. That is the defect working as intended — the type was previously asserting a value is always present when it is not — but it is a compile break on the most widely consumed surface Arc has. No committed proxy in this repository changed, because no TestApps command has an optional property; the churn is entirely downstream.

## The 19-row input space

ARC-33's specs are enumerated from the state machine, not from the diff. Behavior with base domain `myapp.com` and header `X-Tenant-Id`:

| Host | Before | After |
|---|---|---|
| `acme.myapp.com` | `acme` | `acme` |
| `myapp.com` (apex) | `myapp` ❌ | header |
| `www.myapp.com` | `www` | `www` |
| `acme.staging.myapp.com` | `acme` (undefined) | header |
| `10.0.0.5` | `10` ❌ | header |
| `[::ffff:10.0.0.5]` | `[::ffff:10` ❌ | header |
| `acme.myapp.com.` | `acme` | `acme` |
| `ACME.MyApp.COM` | `ACME` ❌ | `acme` |
| `münchen.myapp.com` | `münchen` ❌ | `xn--mnchen-3ya` |
| `acme.localhost`, no base domain | `acme` ❌ | header |

Plus port, empty-host, single-label, punycode-already, and base-domain-is-`www.myapp.com` rows. Restoring the original resolver fails **10 of 23** tenancy assertions.

## One guard was shadowed, and is now pinned

The implementation pass reported three lines it could not falsify. I checked each rather than accepting them.

`IPAddress.TryParse` rejection was reported as dead — but it is **shadowed, not redundant**. It is unreachable only because no *ordinary* base domain lets an address match the suffix rule. With a numeric base domain it is reached, and I confirmed empirically: disabling that one line makes the resolver return **`"10"`** — an address octet — as a tenant id. `and_the_base_domain_is_a_numeric_suffix_of_an_address` now pins it; both of its assertions go red under that mutation.

**Two lines remain genuinely unfalsifiable, and I am naming rather than hiding them:**

- The empty-base-domain short-circuit is provably redundant — with no base domain the suffix becomes `"."`, and normalization trims dots, so no host can match anyway. It survives as an early return and a statement of contract.
- The bracketed-literal rejection is unreachable under any sane configuration, because the address rejection and the suffix rule together already cover every realistic host. Constructing a reaching case needs a base domain containing `]`.

Both are four lines of defense in depth on a tenant-identity boundary, so I kept them. Say the word and they come out.

## Known limitation this PR does not fix

`JavaScriptRuntime.cs` uses `ts.transpile`/`transpileModule` — **syntactic only**. There is no `createProgram` or `getSemanticDiagnostics` anywhere in the repository, so the "compile the generated output" specs structurally cannot catch a type-fidelity defect like ARC-35. The `should_produce_typescript_the_compiler_accepts` assertion proves parseability, not type-correctness; the real coverage here is the textual accessor assertions. Building a semantic-diagnostics harness would catch this whole class of defect and is worth its own change.

## Gates

`dotnet build Arc.slnx` Debug and Release, both `--no-incremental`: **0 warnings, 0 errors**. `Arc.Core.Specs` 1880 passed / 0 failed (plus the 2 added here). `Arc.Specs` 273 passed / 0 failed.

`ProxyGenerator.Specs` — 1104 passed, **3 failed**, all in `when_deserializing_temporal_properties_through_the_generated_proxy` (`TypeError: targetType is not a constructor` from `@cratis/fundamentals` at V8 runtime). **Proven pre-existing**, not accepted on assertion: reverting `Command.hbs` to its original content and rebuilding leaves the same three failing, and that spec exercises `Type.hbs`, which this PR does not touch.

## Not verified

No application was booted against a real multi-tenant host; the resolver evidence is spec-level. No downstream consumer was compiled against the regenerated proxies, so the ARC-35 breakage is reasoned from the emitted type, not observed in a consuming build.

woksin:
Addresses the BLOCK and the HIGH from independent code and security review. `c64f6685`.

## The exploit that was there

Two bugs met. The emptiness check tested the base domain **as written**, while the match used its **normalized** form — and an address literal normalizes to nothing, making the suffix a bare `.`. Separately, normalization could itself *produce* a trailing dot, because `Trim('.')` ran before punycode conversion and IDNA maps U+3002, U+FF0E and U+FF61 onto `.`.

```
UseSubdomainTenancy("192.168.1.10")  +  Host: victim-tenant｡   →   tenant "victim-tenant"
```

The whole attacker-controlled Host header, with no configuration error. Verified across six base domains by the reviewer's probe harness.

**Independently re-verified after the fix.** Reverting the guard to test the raw base domain — exactly the pre-fix code — reddens **both** assertions of `and_the_base_domain_is_an_address_literal`, including `should_not_read_the_host_as_a_tenant`. Restored byte-identically, confirmed with `diff`.

## The fail-open, and why this is now `Removed` + `Security`

`BaseDomain` defaulted to empty, so no host resolved a tenant and every request fell through to `x-cratis-tenant-id` — client-set, stripped nowhere, feeding the Chronicle namespace and the Mongo database name. Before this PR the header was **unreachable** for any multi-label host, so an existing deployment would have converted from host-based to header-based tenancy on upgrade, silently.

It now throws at startup. Since this PR already owns a `major`, failing closed costs nothing.

That also disarms a second trap the code review found: `UseSubdomainTenancy("myapp.com")` bound to the *older* header-only overload, setting `HttpHeader = "myapp.com"` and `BaseDomain = ""` — subdomain tenancy off and the header renamed to something no client sends. Deleting that overload alone would have been worse, silently reinterpreting an existing `UseSubdomainTenancy("X-Tenant-Id")` argument as a base domain. Deleting it **and** requiring a real domain name means that call now fails loudly and says why.

## Where I was wrong, and the implementation was right

I asked for a red-before/green-after spec using `BaseDomain = ""` with `Host: attacker。`. **That spec is green before the fix** — with a literally empty base domain the old short-circuit already returned empty. The exploit needs a base domain that is non-empty but *normalizes* to empty. The address-literal spec above is the true pin, and it was verified red by reproducing the pre-fix code on all three participating lines.

I was also wrong earlier about the `IPAddress.TryParse` rejection. The mutation result I reported was real, but the mechanism was not: `IPAddress.TryParse("0.0.5")` returns **true** (3-part shorthand), so that spec exercised the base-domain operand, not the host. That spec is replaced by `when_creating/and_the_base_domain_is_an_address`, whose summary states what actually happens.

## Decisions on the three unfalsifiable lines

- **`IPAddress.TryParse` — kept.** Dead on the host side given a validated base domain, but load-bearing on the base-domain side: without it `IsDomainName("192.168.1.10")` is true, the configuration is accepted, and every `anything.192.168.1.10` is a tenant. Confirmed — mutating it kills exactly that spec.
- **Bracketed-literal branch — removed.** Confirmed dominated: `IPAddress.TryParse("[::1]")` is true, `WithoutPort` truncates `[x].myapp.com` to `[x]`, and `[` is not LDH.
- **The empty-suffix short-circuit — now genuinely pinned**, and no longer the redundant line it was claimed to be.

## Still not falsifiable — named, not hidden

- `and_the_base_domain_is_a_header_name.should_not_rename_the_tenant_header` dies only to two lines together, because validation throws before the assignment is reached.
- `and_the_base_domain_is_a_label_of_an_unrelated_domain.should_fall_back_to_the_tenant_header` is defended in depth and needs two simultaneous breaks. Its sibling assertion survived **every** mutation that could be constructed, so it was **deleted** rather than shipped as a vacuous pin.
- Four specs carried over from the first pass (`and_host_is_an_ipv4_literal` and three bracket cases) are empirically vacuous — they stay green under the mutations that should kill them. Kept as documentation of intent; no single line falsifies them. Flagging rather than quietly leaving them to read as coverage.

## Not fixed here

`should_produce_typescript_the_compiler_accepts` was renamed to `should_produce_parseable_typescript`, because `ts.transpile` is syntactic only. **Six other occurrences of that name exist outside this diff and overclaim identically** — untouched, worth their own change. The underlying gap stands: no `createProgram`/`getSemanticDiagnostics` anywhere, so the repo structurally cannot catch a type-fidelity defect like #2494.

## Gates

Debug and Release, both `--no-incremental`: **0 warnings, 0 errors**. `Arc.Core.Specs` **1912 passed** (+30). `Arc.Specs` **273 passed**. `ProxyGenerator.Specs` 1104 passed / **exactly 3 failed**, unchanged from the baseline measured before any edit — all in `when_deserializing_temporal_properties_through_the_generated_proxy`, a `@cratis/fundamentals` V8 error in a spec exercising `Type.hbs`, which this PR does not touch.

## Not verified

No application was booted against a real multi-tenant host, and no downstream consumer was compiled against the regenerated proxies — the #2494 breakage is reasoned from the emitted type, not observed in a consuming build.

woksin:
Second round. The tight-scope re-review of `c64f6685` found three more; all fixed in `4c200bcd`.

## 1. The address rejection ran before normalization

`IPAddress.TryParse` tested the **raw** host while the dot-trim and punycode conversion happened after it, and `IsDomainName` only LDH-checked the normalized labels — so the normalized value was never re-tested. Ten spellings of an address were accepted as base domains, and any host below one then resolved a tenant:

`192.168.1.10.` · `.192.168.1.10` · `192.168.1.10..` · `10.0.0.5.` · `127.0.0.1.` · `0.0.5.` · `1.2.3.4.` · `１９２.168.1.10` · `１０.０.０.５` · `192。168。1。10`

`0.0.5` was refused while `0.0.5.` was accepted — the same value one keystroke apart. Normalizing first and rejecting the normalized value closes every spelling at once. All ten now refused; `myapp.com` and `myapp.com.` still accepted, `::1.` now refused by the address check rather than by label-count accident.

**Re-verified independently:** restoring the pre-fix ordering reddens `and_the_base_domain_is_an_address_with_a_trailing_dot` and **only** that spec — the mutation discriminates rather than reddening the whole family.

## 2. The memoized rebuild reinstated the fail-open

The suffix was rebuilt on demand through an **unguarded** path while construction used the guarded one, so a base domain assigned after construction skipped validation and fell back to the client header — the exact HIGH finding, back again.

The rebuild was also dead: the resolver reads `IOptions<T>`, a snapshot, so there was never anything to rebuild for. It is deleted rather than guarded, which removes the second path instead of trying to remember to protect it.

## 3. "Fails at startup" was not true — now it is

It was a first-request failure: `BuildServiceProvider()` succeeded and `ValidateOnBuild` does not catch a factory registration. Fail-closed, but a real user finds the misconfiguration before an operator does. An `IValidateOptions<ArcOptions>` now runs under the `ValidateOnStart()` already wired on all three entry points — one registration line. The specs prove start-not-build: they call `.Build()` successfully and catch `BaseDomainIsNotADomainName`, unwrapped, around `StartAsync()`.

## Corrections to my own earlier comment

- I cited `should_not_read_the_host_as_a_tenant` as a paired positive when reporting the exploit mutation. **It is not independent evidence** — it is logically implied by its `ShouldEqual(HeaderTenantId)` sibling and can never go red alone. Both did redden, but that is one observation, not two. Six assertions of that shape are now annotated as restatements rather than left to read as pins; the re-review found three of them that were not in my list.
- My previous claim that startup validation was in place was wrong, per item 3.

## Two specs the fix made vacuous, handled rather than left

`and_no_base_domain_is_configured` and `and_the_base_domain_is_an_address_literal` both wrote `BaseDomain` *after* construction and depended on the rebuild to observe it. With the rebuild gone they pass vacuously — the state they describe cannot exist. The first is replaced by `and_the_base_domain_is_changed_after_the_resolver_was_created`, which is the direct pin for item 2; the second is deleted, its coverage now carried by `when_creating/and_the_base_domain_is_an_address`, the new trailing-dot spec, and the startup specs.

`and_base_domain_is_the_www_host` was red for the same reason and is repaired. It has no single-line break it dies to independently, which is stated in its summary, and `and_host_is_a_tenant_below_the_www_base_domain` was added to carry the evidence.

## What the re-review confirmed clean

Regex anchoring is `\A`/`\z`, not `^`/`$` — every line-break input rejected. Trimming after punycode introduces no new collisions; the observed ones are intended canonicalizations to the same tenant. **3.2M concurrent resolves across 8 threads with a concurrent config flipper produced zero anomalies.** No remaining call site or documentation example uses the deleted overload.

## Known and unenforced

`IsDomainName` accepts a public suffix (`co.uk`), 300 labels, and all-numeric labels. There is no public-suffix list and no 253-byte cap. Documented as convention in `resolvers.md`, not enforced — flagging so it is a decision rather than a surprise.

## Gates

Debug and Release, `--no-incremental`: **0 warnings, 0 errors**. `Arc.Core.Specs` **1914 passed**. `Arc.Specs` **273 passed**. `ProxyGenerator.Specs` 1104 passed / exactly **3** failed — unchanged from the pre-change baseline, all in `when_deserializing_temporal_properties_through_the_generated_proxy`, which exercises `Type.hbs` and is untouched here.

woksin:
Round-two re-review: **MERGE**, no correctness or security defect. Both of its low-severity notes are now addressed in `334bdcfb`.

## What round two actually proved

Not by reasoning — by running inputs against the built `Cratis.Arc.Core.dll`:

- Every one of the ten previously-accepted address spellings is now refused, plus ~40 more: decimal/octal/hex integer forms (`2130706433`, `0177.0.0.1`, `0x7f.1`), every IPv6 form including zone ids and bracketed-with-port, fullwidth and superscript and circled digits, and digits separated by soft hyphen, ZWSP or ZWJ.
- A **3,045-input corpus**: zero accepted base domains whose normalized value parses as an address, zero bare-dot or dotless suffixes, and `Normalize` idempotent on all of them.
- **300,000 fuzzed hosts**: zero tenants outside `\A[a-z0-9]([a-z0-9-]{0,61}[a-z0-9])?\z`, zero throws.
- **200,000 fuzzed base domains × 5 attacker hosts**: zero wildcard suffixes.

The reordering is structurally complete — the second `Trim('.')` runs before the address test, so `Normalize` cannot return an address-parseable string.

## The false-positive check, which was the thing I most wanted answered

**No false positive**, verified across all three entry points with real hosts. A non-subdomain application with an empty `BaseDomain` starts normally; only `ResolverType == Subdomain` with a missing or non-domain base domain refuses. The exception is `BaseDomainIsNotADomainName` **unwrapped** everywhere, never repackaged into `OptionsValidationException`. `TryAddEnumerable` yields exactly one registration after two `AddCratisArcCore()` calls — it can be skipped, never dropped. It also fires on a bare `IOptions<ArcOptions>.Value` resolution, so the testing path, which has no `ValidateOnStart`, is fail-closed too.

It also corrected a detail I had wrong: `ArcApplicationBuilder` fails at **`Build()`**, not `StartAsync()`, because `Build()` resolves `.Value` — earlier than I claimed, not later.

## The two notes, both taken

**A spec now names the most dangerous line in the validator.** The `ResolverType == Subdomain` condition was caught only incidentally, by a spec whose name (`and_the_environment_is_production`) gives no hint that it is what stands between us and refusing to start every header-, query- and claim-tenanted application. `and_tenancy_is_resolved_from_the_header` names it. **Verified: making the guard unconditional reddens exactly that spec and nothing else.**

**`TenancyOptionsValidator` is now in the body's Added section** — it is new public API on a `major`.

## Confirmed on the specs that were replaced

`and_the_base_domain_is_changed_after_the_resolver_was_created` discriminates both ways: under the old memoized path it goes red via the unguarded rebuild, and under a *guarded* rebuild it goes red by throwing. Deleting `and_the_base_domain_is_an_address_literal` drops nothing — the bare-dot suffix state is now unreachable, and its coverage is carried by three other specs. The new trailing-dot spec is genuinely red pre-fix, because `IPAddress.TryParse("192.168.1.10.")` returns **false** — measured, not assumed, which is exactly why the old ordering let it through.

## Gates

Debug and Release, `--no-incremental`: **0 warnings, 0 errors**. `Arc.Core.Specs` **1915 passed**. `Arc.Specs` **273 passed**. `ProxyGenerator.Specs` unchanged at exactly 3 known pre-existing failures.

woksin:
**Split, to avoid a major.** `a848a1cb` reverts the generated-proxy nullability change; this PR is now tenancy only and relabelled `minor`. Retitled to match.

**#2494 is not fixed by this PR and stays open.** The defect is real — a nullable command property still generates accessors that claim a value is always present. It returns on its own branch so the break can be taken deliberately rather than as a passenger on a security fix.

## Why that was the only thing forcing a major

Generated proxies are the most consumed surface Arc has, and every application regenerates them on a Release build. Widening an optional property's getter from `T` to `T | undefined` makes `const x: string = command.description` fail to compile with no code change on the consumer's side — a source break, even though the runtime behaviour, wire format and `PropertyDescriptor` are all unchanged.

Reverted **additively** rather than by rewriting the branch, so the history keeps the record of what was tried and why it was pulled.

## Why the remainder is defensible as `minor`

Two things in here are breaking on a strict reading, and I would rather name them than have them found later:

- **A public overload was deleted.** `UseSubdomainTenancy(string? fallbackHeaderName = null)` is gone. It has **zero call sites** across Arc, AuthProxy, Chronicle, Components and the consuming application that reported this. A caller passing a header name where a base domain now belongs does not silently misbehave — it fails at startup with a message naming the problem.
- **Subdomain tenancy without a base domain now throws at startup.** It previously started and resolved every request's tenant from a client-settable header. Refusing to start is strictly safer than what it replaced, and it only fires for a configuration that was already insecure.

Neither is a frequent consumer API, and both fail loudly rather than quietly. That is the basis for `minor`; flagging it explicitly so it is a decision on the record rather than an oversight.

## Gates after the revert

`Command.hbs` is byte-identical to the merge base — verified, not assumed. Debug and Release, both `--no-incremental`: **0 warnings, 0 errors**. `Arc.Core.Specs` **1915 passed**. `Arc.Specs` **273 passed**. The two proxy specs that only existed to cover the reverted change were removed with it.

PR #2519, opened 2026-08-11
Carry nullability into a generated command's property accessors

woksin:
## Fixed

- A generated command proxy declares an optional property as optional, so its accessors carry `| undefined` instead of claiming a value is always present (#2494)

### DIFF
--- a/Source/DotNET/Tools/ProxyGenerator.Specs/Scenarios/for_ProxyGeneration/CommandWithOptionalProperties.cs
+++ b/Source/DotNET/Tools/ProxyGenerator.Specs/Scenarios/for_ProxyGeneration/CommandWithOptionalProperties.cs
@@ -1,0 +1,36 @@
+// Copyright (c) Cratis. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Cratis.Arc.ProxyGenerator.Scenarios.for_ProxyGeneration;
+
+/// <summary>
+/// A command mixing required and optional properties, for verifying that the generated proxy keeps the nullability of
+/// every property.
+/// </summary>
+public class CommandWithOptionalProperties
+{
+    /// <summary>
+    /// Gets or sets the required name.
+    /// </summary>
+    public string Name { get; set; } = string.Empty;
+
+    /// <summary>
+    /// Gets or sets the optional description.
+    /// </summary>
+    public string? Description { get; set; }
+
+    /// <summary>
+    /// Gets or sets the optional value.
+    /// </summary>
+    public int? Value { get; set; }
+
+    /// <summary>
+    /// Gets or sets the required tags.
+    /// </summary>
+    public IEnumerable<string> Tags { get; set; } = [];
+
+    /// <summary>
+    /// Gets or sets the optional labels.
+    /// </summary>
+    public IEnumerable<string>? Labels { get; set; }
+}
--- a/Source/DotNET/Tools/ProxyGenerator.Specs/Scenarios/for_ProxyGeneration/when_generating_command_with_optional_properties.cs
+++ b/Source/DotNET/Tools/ProxyGenerator.Specs/Scenarios/for_ProxyGeneration/when_generating_command_with_optional_properties.cs
@@ -1,0 +1,67 @@
+// Copyright (c) Cratis. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Cratis.Arc.ProxyGenerator.Scenarios.Infrastructure;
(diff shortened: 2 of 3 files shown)


woksin:
**Draft deliberately — this is the breaking half of #2517, split out so the major can be taken on purpose.**

It shipped once, on the branch that became v21.6.0, and was reverted before merge so the subdomain tenancy security fix could reach consumers on a `minor`. This is the same change, re-applied on its own branch. Nothing here is urgent; merge it whenever a major suits.

## Why it is a major

`Templates/Command.hbs` already knew which properties were nullable — it used that on the wire descriptor — but the backing field and accessors ignored it. So a nullable command property generated as non-nullable, and the proxy's type asserted a value is always present when it is not.

Generated proxies are the most consumed surface Arc has, and every application regenerates them on a Release build. A consumer writing `const x: string = command.description` starts getting TS2322 **with no code change of their own**. That is a source break, and it is the whole reason this cannot ride along on a minor.

The honest counter-argument, for whoever decides: **runtime behaviour is identical.** The wire format is unchanged, `PropertyDescriptor(name, type, isNullable)` was already emitted correctly and is untouched, and the server binds JSON against the C# record either way. Applications without `strictNullChecks` see nothing at all. Applications with it get an error pointing at a latent bug they already had. So this narrows a claim rather than changing what runs.

## What this cannot prove, and why it matters here

`JavaScriptRuntime.cs` uses `ts.transpile`/`transpileModule` — **syntactic only**. There is no `createProgram` or `getSemanticDiagnostics` anywhere in the repository, so the "compile the generated output" specs structurally cannot catch a type-fidelity defect. That is precisely why this defect survived.

The assertion formerly named `should_produce_typescript_the_compiler_accepts` was renamed to `should_produce_parseable_typescript` in #2517, because the old name overclaimed. **Six other occurrences of that name remain elsewhere in the repository and overclaim identically** — untouched, worth their own change.

Building a semantic-diagnostics harness would catch this whole class of defect and would be far more valuable than this one fix. Worth considering before or alongside merging.

## Evidence

Reverting the template makes all 7 optional-property assertions fail, while the 6 required-property baselines, the `PropertyDescriptor` assertions and the TypeScript-parse assertion stay green — so the change is pinned from both sides rather than only asserting the new shape.

## Gates

`dotnet build Arc.slnx -c Release --no-incremental`: **0 warnings, 0 errors**. `ProxyGenerator.Specs`: 1104 passed, **exactly 3 failed** — `should_hand_back_the_time_of_day_the_server_sent`, `should_not_shift_the_calendar_day_west_of_utc`, `should_hand_back_the_calendar_date_the_server_sent`, all in `when_deserializing_temporal_properties_through_the_generated_proxy`. Confirmed pre-existing and unrelated: that spec exercises `Type.hbs`, which this PR does not touch, and the same three fail on an unmodified tree.

woksin:
Relabelled `major` → `patch` and taken out of draft.

The reasoning, recorded so the decision is inspectable later: the generated type was simply **wrong**. It asserted a value is always present for a property the C# record declares nullable, and the template already knew that — it used the same `IsNullable` on the wire descriptor. Correcting it restores behaviour that should have been there from the start rather than changing anything.

Nothing observable moves:

- **Wire format unchanged.** `PropertyDescriptor(name, type, isNullable)` was already emitted correctly and is untouched, so the request body is byte-identical and the server binds against the same C# record either way.
- **No public C# surface changed.** Nothing added, removed or re-signatured.
- **Applications without `strictNullChecks` see nothing at all.**

The one consumer-visible effect, stated plainly rather than buried: an application **with** `strictNullChecks` that reads an optional command property into a non-nullable position will get TS2322/TS2345 on regeneration. That error is the fix working — it points at a latent bug the application already had, which the proxy's incorrect type was hiding. Under this repository's convention (`patch` — bug fixes) that is a fix surfacing, not a breaking change.

## Review coverage

The template change was examined by the independent code review on #2517, where it originally shipped: branches confirmed correct for scalar, value-type, enumerable, concept, enum and nested properties — `isNullable` is computed before the enumerable unwrap, so every kind routes identically — and no invalid-TypeScript shape was found. Element nullability (`IEnumerable<string?>`) is still dropped; that is pre-existing and out of scope.

## Still true, and still worth doing separately

`JavaScriptRuntime.cs` uses `ts.transpile`/`transpileModule` — syntactic only, with no `createProgram` or `getSemanticDiagnostics` anywhere in the repository. The "compile the generated output" specs therefore **structurally cannot** catch a type-fidelity defect, which is exactly why this one survived. The real coverage here is the textual accessor assertions. A semantic-diagnostics harness would catch this whole class and is worth more than this fix.

Six assertions elsewhere still named `should_produce_typescript_the_compiler_accepts` overclaim identically; only the one in scope was renamed.

Issue #2521, opened 2026-08-11
Screenplay generation crashes on any multi-project application with specifications: GetSemanticModel is called on a tree from a referenced project's compilation


woksin:
`cratis screenplay generate` produces **no document, no diagnostics and no partial result** against a real multi-project Arc application:

```
$ cratis screenplay generate Source/Core/Core.csproj
Error: SyntaxTree is not part of the compilation (Parameter 'syntaxTree')
```

Reproduced against Cratis Studio's `Core.csproj` and against its `Studio.slnx`, with both `Cratis.Arc.Screenplay` 21.2.0 (via the CLI) and `origin/main` at 21.6.0 (via `Source/DotNET/Screenplay.EndToEnd`). Identical crash, identical line.

## Crash site

`Source/DotNET/Screenplay/Analysis/Specifications/GeneratedIdentities.cs:117`

```csharp
return symbol.DeclaringSyntaxReferences
    .Select(_ => ValueOf(_.GetSyntax()))
    .OfType<ExpressionSyntax>()
    .Any(value => Yields(value, compilation.GetSemanticModel(value.SyntaxTree), followed));
```

Path in: `SpecificationCatalog.Read` → `SpecificationReader.Read:88` → `SpecificationStepReader.ReadGiven:57` → `SpecificationValues.Add:136` → `GeneratedIdentities.Yields:50` → `Holds:114`.

## Mechanism

`MSBuildWorkspace.OpenProjectAsync` opens the project **and its project references**, and `GetCompilationAsync` wires those references as **`CompilationReference`s rather than `MetadataReference`s**. Symbols from a referenced project are therefore *source* symbols carrying real `DeclaringSyntaxReferences` that point at syntax trees owned by a **different `Compilation`**.

`GeneratedIdentities.Holds` follows any field or property a specification names back to its initializer, then asks the *root* compilation for a semantic model on that foreign tree. `Compilation.GetSemanticModel` throws for any tree it was not built from.

Measured against Studio's `Core.csproj`:

```
syntax trees in compilation:                    1380
CompilationReference (source symbols):             9
    Admin.Contracts, Infrastructure, Lobby.Contracts, PrimeReact,
    Scene, Shared, Catalog, Catalog.Contracts, Core.Contracts
PortableExecutableReference (metadata only):     487

field/property symbols whose declaring tree is FOREIGN:  264 distinct files
    Admin.Organizations.AnnouncementBody.NotSet
[...]

PR #2522, opened 2026-08-11
Read a scenario's values through the project that declares them

woksin:
## Changed

- Generating from a single project now reads every project it references as well, instead of only the project named. An application whose events, commands and concepts live in contracts projects below the one named no longer generates a document that leaves all of them out (#2521)

## Fixed

- Generating a Screenplay document no longer fails with `SyntaxTree is not part of the compilation` on an application written as several projects. A scenario naming a value another project declares - a sentinel identity, a well known identity, a member of an enumeration - is now read through the project that actually holds it, so it is recognized for what it is instead of taking generation down with it (#2521)
- A value a scenario states that no handed-over project declares is now left unrecovered and reported, rather than throwing (#2521)

### DIFF
--- a/Source/DotNET/Screenplay.EndToEnd/ProjectCompilation.cs
+++ b/Source/DotNET/Screenplay.EndToEnd/ProjectCompilation.cs
@@ -20,7 +20,9 @@ namespace Cratis.Arc.Screenplay.EndToEnd;
 /// construction. Two shipped that way.
 /// <para>
 /// A solution is read as every project it holds that could hold part of the application, which is what an application
-/// written as several projects really is. Whether a project could is asked of what it can see rather than of what it
+/// written as several projects really is. A single project is read the same way, because opening one opens everything
+/// it references too - an application is no less written as several projects for having been named by its topmost one.
+/// Whether a project could hold part of it is asked of what it can see rather than of what it
 /// is called: an artifact is declared with an attribute the framework ships, so a project resolving neither the Arc
 /// nor the Chronicle one cannot declare a single thing the document is made of. A Roslyn analyzer beside the
 /// application is exactly that project, and reading its <c>netstandard2.0</c> compilation in as though it were part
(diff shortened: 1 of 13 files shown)


woksin:
## Reviewer context

### Mechanism

`MSBuildWorkspace.OpenProjectAsync` / `OpenSolutionAsync` wire project references as **`CompilationReference`s, not `MetadataReference`s**. A type a referenced project declares is therefore a *source* symbol whose `DeclaringSyntaxReferences` point at trees owned by a **different** `Compilation`. `Compilation.GetSemanticModel(tree)` throws `ArgumentException` for any tree it does not own, so following such a member is a crash rather than a wrong answer.

Measured against `Studio/Source/Core/Core.csproj`: 1380 trees, 9 `CompilationReference`s, **264 distinct files** holding field/property symbols foreign to the root compilation — `NotSet` sentinels and enum members in `*.Contracts` projects, i.e. the ordinary house architecture.

`SemanticModels` (`Analysis/SemanticModels.cs`) already existed for exactly this, its own remarks describe the crash verbatim, and `WholeApplication` already constructs it — but only `ProducesReader` consumed it. This routes the rest through it.

### Why handing over more compilations was not enough on its own

`SpecificationCatalog.Read` is called from `CompilationAnalysis.Specifications` with the **per-project** `Compilation`, so `GeneratedIdentities` was single-compilation even on the solution path where every compilation *was* already handed over. That is why the solution target crashed too, and why the `SemanticModels` routing — not the loader change — is the load-bearing half.

Verified by reverting the loader change alone: the single-project target still produced a document (8451 lines, 1 project). This matters for the cli, whose single-`.csproj` path has the same loader shape as Arc's did — after this lands it degrades to SP0040 instead of crashing, without any cli change.

### Sites

Confirmed crashing: `Specifications/GeneratedIdentities.cs:117`. Latent and fixed alongside it: `SpecificationStepReader`, `SpecificationReader`, `SpecificationOutcomeReader`, `Validation/ValidationReader`, `Constraints/ConstraintReader`, `Projections/FluentProjectionReader`, `Reactors/ReactorReader` — each reachable the moment a base context, validator, fluent projection or reactor handler is inherited from a project below. `Policies/PolicyRegistrations.cs:55` is safe by construction (it iterates `compilation.SyntaxTrees`) and is left alone.

Degradation follows the pattern `ProducesReader` established: say nothing rather than guess. The value is then reported as one nothing was recovered from (SP0040) instead of taking the run down.

`ArtifactReaders.For(compilation, catalog, paths, whole)` lost its `compilation` parameter — every reader it composed now takes `SemanticModels`, so nothing was left for it to do.

### Why the specification is shaped the way it is

`ProjectCompilation`'s own remarks note that two defects have already shipped through this hole; this was the third. The class is invisible to `Screenplay.Specs` only when every compilation is built from source strings in isolation — the multi-project harness (`Analyzed.Project` + `ToMetadataReference()`) already produces real `CompilationReference`s. What was missing was a spec putting a **scenario in one project that names a member declared in another**.

`a_scenario_stating_an_identity_a_sibling_project_declares` does that, and asserts on the **values** rather than on not crashing: with the fix the cross-project identity is still recognized as made on the spot and therefore still left out of the document. A guard that merely declined to read across the boundary would compile, stop the crash, and **fail** this spec — which is the point.

Against unfixed `main` all 5 facts fail with the production stack trace at `GeneratedIdentities.cs:117`.

### Verification

- `Screenplay.Specs` — 1228 passed, 0 failed (232 spec files).
- `Screenplay.EndToEnd` against `Studio/Source/Core/Core.csproj`: before, `ArgumentException` with no document and no diagnostics; after, 10 projects and an 8582-line document.

### Out of scope, seen in passing

The Studio document now reaches the language compiler and is rejected by 27 warnings — unknown types `KeyValuePair`, `UIElement`, `IRule`, `ICaptureSource`, `ICaptureOutput` on dictionary- and interface-typed properties. That is a separate generator gap that the crash was previously hiding, not a regression from this change.

PR #2523, opened 2026-08-11
Carry every projection a slice declares

woksin:
## Changed

- A generated slice now carries every projection it declares, instead of the first one and a diagnostic for the rest. An application whose slice builds several read models - the one a screen binds to and the one a command reads to decide - now gets all of them
- A reducer's read model now stands beside the slice's projections instead of competing with them, so a slice mixing projections and reducers no longer loses most of its read side
- `SP0015` no longer reports a second projection in a slice; it is left for the constructs within a projection that the language genuinely has no word for
- `SP0020` now says what is actually lost - the read model is carried as a projection over the events the reducer observes, and only the fold itself goes unstated

### DIFF
--- a/Source/DotNET/Screenplay.Specs/Library/LibraryAuthors.cs
+++ b/Source/DotNET/Screenplay.Specs/Library/LibraryAuthors.cs
@@ -36,7 +36,7 @@ public static class LibraryAuthors
             ],
             [new EventModel("AuthorRegistered", [Declare.Property("Name", "AuthorName")], ["audit", "authors"])],
             [],
-            null,
+            [],
             [],
             [new UniquePropertyConstraintModel("UniqueAuthorName", "Name", "AuthorRegistered")]);
 
@@ -61,16 +61,18 @@ public static class LibraryAuthors
                     [],
                     null)
             ],
-            new ProjectionModel(
-                "Library.Authors.Listing.AuthorProjection",
-                "Author",
-                "event-log",
-                ProjectionAutoMapMode.Enabled,
-                false,
-                ProjectionScopeModel.Empty with
-                {
-                    From = [new(["AuthorRegistered"], "$eventSourceId", null, Declare.Map(("name", "name")))]
-                }),
+            [
+                new ProjectionModel(
+                    "Library.Authors.Listing.AuthorProjection",
+                    "Author",
+                    "event-log",
+                    ProjectionAutoMapMode.Enabled,
+                    false,
+                    ProjectionScopeModel.Empty with
+                    {
+                        From = [new(["AuthorRegistered"], "$eventSourceId", null, Declare.Map(("name", "name")))]
+                    })
+            ],
             [],
             []);
 }
--- a/Source/DotNET/Screenplay.Specs/Library/LibraryInventory.cs
+++ b/Source/DotNET/Screenplay.Specs/Library/LibraryInventory.cs
@@ -56,7 +56,7 @@ public static class LibraryInventory
                     [])
             ],
             [],
-            null,
+            [],
             [],
             []);
 
@@ -73,23 +73,25 @@ public static class LibraryInventory
             [],
             [],
             [new QueryModel("AllBooks", Declare.Many("Book"), null, [], null)],
-            new ProjectionModel(
-                "Library.Inventory.Listing.BookProjection",
-                "Book",
-                "event-log",
-                ProjectionAutoMapMode.Enabled,
-                false,
-                ProjectionScopeModel.Empty with
-                {
-                    From =
-                    [
-                        new(
-                            ["BookAddedToInventory"],
-                            "$eventSourceId",
-                            null,
-                            Declare.Map(("title", "title"), ("available", "count")))
-                    ]
-                }),
+            [
+                new ProjectionModel(
+                    "Library.Inventory.Listing.BookProjection",
+                    "Book",
+                    "event-log",
+                    ProjectionAutoMapMode.Enabled,
+                    false,
+                    ProjectionScopeModel.Empty with
+                    {
+                        From =
+                        [
+                            new(
+                                ["BookAddedToInventory"],
+                                "$eventSourceId",
+                                null,
+                                Declare.Map(("title", "title"), ("available", "count")))
+                        ]
+                    })
+            ],
             [],
             []);
 }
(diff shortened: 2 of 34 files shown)


woksin:
## Reviewer context

### What was actually wrong

`SliceSyntax` has taken `IEnumerable<ProjectionSyntax>` since Screenplay 2.0.0, and `SliceSyntaxBuilder.BuildProjections` already yielded a list — of at most one element, with a comment saying "widening the analysis is the only thing left to do to carry them all". This is that widening: the emission side is unchanged in shape.

`SliceModel.Projection` → `Projections`, `SliceContents.Projection` → `Projections`, and the two turn-away sites (`SliceArtifactReader.Assign`, `SliceUnion.OneProjection`) are gone. The union now joins projections through the same `Once` helper every other artifact kind uses, so two projects declaring the same projection identifier in one slice report `RepeatedDeclarationAcrossProjects` exactly as a repeated command or event does.

### The reducer was the expensive half

`ReducerReader.Read` synthesises a `ProjectionModel` from the events a reducer observes. With one slot per slice, every reducer competed with every real projection, and which read model survived was decided by catalogue order. `cli/assets/demo-store` (2 projections + 2 reducers) is the measurement:

| | before | after |
|---|---|---|
| read models carried | 1 of 4 (`Book`) | 4 of 4 (`Book`, `BorrowedBook`, `Member`, `OverdueBook`) |
| diagnostics | `SP0015 x3`, `SP0020 x2`, `SP0023 x1` | `SP0020 x2`, `SP0023 x1` |
| document | 36 lines | 57 lines |

Both documents read back clean through `ScreenplayCompiler`.

### On the two diagnostics

Neither code is deleted. `SP0015` had 8 emission sites; the 2 second-projection ones are gone and the other 6 — in `FluentMappings`, `FluentBlockReader`, `FluentScopeReader` (×3) and `ModelBoundScopeReader` — are genuinely unmappable constructs *within* a projection and still fire. Its XML remarks described the removed behavior and now describe what is left.

`SP0020` still fires once per reducer, because the fold really is unstated. Its old text — "which Screenplay has no counterpart for" — was false against Screenplay 2.x, which has a `reducer` construct; the new text says what is true today: the read model is carried as a projection over the observed events and only the fold goes unsaid.

**Deferred:** emitting a reducer as `reducer` syntax rather than as a synthesized projection. `ReducerSyntax` and `SliceSyntax.Reducers` exist in 2.1.0, so this is available — but it needs an emission builder, and the fold body still needs a `file` reference, so `SP0020` narrows rather than closes either way. Holding it for the follow-up PR alongside `observable` and `produces … for` keeps this one to the change that unblocks the read side.

### Specs

`a_slice_declaring_a_second_projection` asserted the defect — that only the first projection survives and that the message says "a slice may declare at most one". It is replaced by `a_slice_declaring_several_projections`, which asserts both are carried and nothing is reported.

`a_slice_declaring_a_projection_and_a_reducer` is new and covers the shape that actually cost the demo store its read side: a reducer and a projection in one slice, asserting both read models land and that `SP0020` still fires for the fold.

### Verification

- `Screenplay.Specs` — 1226 passed, 0 failed.
- `Screenplay.EndToEnd` against `cli/assets/demo-store/DemoStore.csproj` — the table above.

woksin:
## Addendum — a defect this PR surfaced, and the fix for it

Running the combined wave against a real application (`Studio/Source/Core/Core.csproj`, 10 projects) caught something `Screenplay.Specs` and CI both missed: **Screenplay holds one builder per read model across the whole document**, not per slice. Two Studio slices both project into `Domain`, and once every projection was carried the emitted document stopped compiling:

```
generation error SP0034: The generated document did not compile - 1 error(s), the first being
'Read model 'Domain' is built more than once - a projection or a reducer builds it, and only one of them'
```

CI was green on the previous commit of this PR — Arc's own `screenplay-end-to-end` sample has no duplicate read model, so nothing in the pipeline reached it. Only a real multi-project application did.

`SliceUnion.OneBuilderPerReadModel` now applies that rule once the slices are joined, keeping the first builder in namespace order and reporting the rest as `SP0015`. So the old "at most one projection per slice" rule had a real half after all — it was just keyed on the wrong thing. Dropping every projection after the first was too broad and cost read models that were never in conflict; dropping none at all writes a document nothing can read.

`two_slices_building_one_read_model` covers it.

### Numbers after the fix

| | before the wave | after |
|---|---|---|
| Studio projections carried | 53 | 62 |
| Studio `SP0034` generation errors | — | none (1 × `SP0015` instead) |
| demo store read models | 1 of 4 | 4 of 4, document reads back clean |

The 27 diagnostics the Studio document still reports back are all the pre-existing `Unknown type` class (`KeyValuePair` ×17, `UIElement` ×5, `ICaptureSource`/`ICaptureOutput` ×2 each, `IRule` ×1) on dictionary- and interface-typed properties — the same 27 as before this PR, unrelated to it.

`Screenplay.Specs` is 1230 passed, 0 failed; Debug and Release both build 41 projects with 0 warnings.

woksin:
## On the label — `minor`, but it is a judgment call worth confirming

`SliceModel.Projection` (`ProjectionModel?`) became `SliceModel.Projections` (`IEnumerable<ProjectionModel>`), and `SliceModel` is a public positional record — so this is source and binary breaking for anyone who constructs or deconstructs one. By the letter of the repo's rule that is `major`.

I labelled it `minor` because no consumer touches it: the cli's Screenplay commands use only `ScreenplayGenerator`, `ScreenplayOptions` and `ScreenplayDiagnostic`, and Studio consumes the emitted `.play` document rather than Arc's model types. `SliceModel` is generator-internal in practice, and the observable change for a user of Arc is purely additive — the document carries more than it did.

Relabel to `major` before merging if you would rather hold the line on the letter of the rule. Same question applies to `QueryModel` in the follow-up PR, though that one only gains a trailing optional parameter and so is source compatible.

woksin:
Keeping **`minor`**, on precedent rather than on the letter of the rule.

#2469 made the same class of change to the same package — `AuthorizeSyntax` lost `.Policies` and `SliceSyntax` lost `.Projection` — and was deliberately labelled `minor` on the owner's call. The reasoning recorded then applies unchanged here:

- `Cratis.Arc.Screenplay` has one real consumer, the CLI, which is maintained in-house and bumped in lockstep.
- A `major` on this package forces one across the whole Arc surface, `Cratis.Arc.Core` included, for a change confined to the Screenplay generator.
- Residual risk accepted knowingly; no known outside consumer.

Your flag was the right thing to raise — `SliceModel.Projection` → `Projections` is a positional-record break and `major` by the letter. Recording the precedent here so the next person does not have to rediscover it, and so nobody "corrects" the label without reading this first.

PR #2524, opened 2026-08-11
State that a query keeps answering

woksin:
## Added

- A query returning `ISubject<T>` is now marked `observable` in the generated document, so a live read that keeps pushing as the read model changes is no longer written as though it answers once

### DIFF
--- a/Source/DotNET/Screenplay.Specs/for_ApplicationModelAnalyzer/when_analyzing/a_query_that_keeps_answering.cs
+++ b/Source/DotNET/Screenplay.Specs/for_ApplicationModelAnalyzer/when_analyzing/a_query_that_keeps_answering.cs
@@ -1,0 +1,59 @@
+// Copyright (c) Cratis. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Cratis.Arc.Screenplay.Analysis;
+using Cratis.Arc.Screenplay.Model;
+
+namespace Cratis.Arc.Screenplay.for_ApplicationModelAnalyzer.when_analyzing;
+
+/// <summary>
+/// Handing back a subject is the whole of how a query says it is a live read - there is no attribute saying so, and
+/// the caller subscribes rather than asks. Every other wrapper really is only how a result arrives, so peeling them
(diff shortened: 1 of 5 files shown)


woksin:
## Reviewer context

### The shape of the loss

`QueryReturnTypes` lists `ISubject<T>` among the wrappers it peels, alongside `Task`, `ValueTask`, `IQueryable` and `ActionResult`. For the other four that is right — they say how the result arrives, never what it is. `ISubject<T>` is different: it is the *only* thing that says a query is a live read. There is no attribute and no parameter, and the caller subscribes rather than asks. So peeling it lost a fact about the application with **no diagnostic at all** — the one silent drop, against a generator that otherwise reports everything it cannot carry.

`QuerySyntax.IsObservable` has existed since Screenplay 2.0.0 as a trailing optional, so nothing on the language side needed to change.

`QueryReturnTypes.IsObservable` mirrors `IsPagedByTheHost` exactly — walk the single-type-argument chain, match, return. It is asked of the raw return type before `Unwrap` runs, so it is independent of the peeling order.

### Specs

Nothing in `Screenplay.Specs` referenced `ISubject` at all, which is why the drop survived — this is the vacuity that let it ship. `a_query_that_keeps_answering` declares `System.Reactive.Subjects.ISubject<T>` as source (the spec compilation has no System.Reactive reference, and `Matches` resolves by fully qualified metadata name) and asserts both directions: the subject-returning query is observable, the plain one is not, and the return type is still recovered through the subject.

### Verification

- `Screenplay.Specs` — 1228 passed, 0 failed.

### On `produces … for`, which is not in this PR — the brief for it is wrong

I was told the analysis for `produces … for` was "already done", because `ProducesReader` detects the event-source-id result in order to report `SP0013`. It does not. `HandlerBodies.YieldsEventSourceId` is a pure **return-type shape** check — it asks whether the handler returns a tuple with more than one element of which one is an event. It never identifies *which* element is the event source id, and never touches the expression.

`ProducesSyntax.For` is an `ExpressionSyntax?`, so emitting it needs the actual expression the handler yields, recovered and converted the way `ProducesMappingReader` recovers a mapping source. That is real analysis work, not a wiring change, and `SP0013` cannot be retired until it exists. I have left it out rather than half-build it.

PR #2525, opened 2026-08-11
Name the application after the project that was asked for

woksin:
## Fixed

- Generating from a single project again names the document after that project, instead of falling back to the neutral default. Handing over the projects it references made the application several projects, and several projects are named by none of them (#2521)

### DIFF
--- a/Source/DotNET/Screenplay.EndToEnd/LoadedApplication.cs
+++ b/Source/DotNET/Screenplay.EndToEnd/LoadedApplication.cs
@@ -1,0 +1,18 @@
+// Copyright (c) Cratis. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.CodeAnalysis;
+
+namespace Cratis.Arc.Screenplay.EndToEnd;
+
+/// <summary>
+/// Represents what reading a project or a solution yielded.
+/// </summary>
+/// <param name="Compilations">The compilations to generate from, empty when nothing yielded one.</param>
+/// <param name="Name">The name of the project that was asked for, or <see langword="null"/> when a solution was.</param>
+/// <remarks>
+/// An application read as several projects is named by none of them, so the generator offers no name and the caller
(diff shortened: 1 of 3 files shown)


woksin:
## Reviewer context

Follow-up to #2522, which merged before I caught this.

Handing over the projects a named project references (that PR's loader change) makes the application **several** projects. `AnalyzedCompilations.NameOf` deliberately offers no name in that case — "several projects are the application together and none of them names it, picking one would put the name of a layer where the name of the application belongs" — so `ScreenplayOptions.DefaultName` took over and the document's first line changed:

```
before #2522:   domain Studio
after  #2522:   domain Application
this PR:        domain Studio
```

The design is right; the caller is supposed to supply the name when there are several projects, and `Screenplay.EndToEnd` was not supplying one. The project the caller pointed at *is* the caller having already said which one the application is, so `ProjectCompilation` now hands it back (`LoadedApplication`) and `Program` passes it as `Domain`. A solution target names nothing this way, because pointing at a solution points at all of it equally — `Name` is null and the existing fallback applies unchanged.

### Scope

`Cratis.Arc.Screenplay.EndToEnd` is `IsPackable=false` — an internal check, not a shipped package — so no consumer saw the regression. **The cli is the one that matters**: its single-`.csproj` path still hands over one compilation, so it is unaffected today. If the cli adopts the same loader change, it has to supply the domain the same way or every generated document will be named `Application`.

### Verification

- `Screenplay.Specs` — 1228 passed, 0 failed.
- Both apps the `screenplay-end-to-end` job generates from read back clean and are named correctly: `domain AspNetCore`, `domain Chronicle`.
- `Studio/Source/Core/Core.csproj` — 10 projects, 8582 lines, `domain Studio`.

### A gap worth knowing about

The `screenplay-end-to-end` job generates from `TestApps/AspNetCore` and `TestApps/Chronicle`. Both are single applications whose project references are Arc's own framework projects, so neither exercises a multi-project *application* — which is why the crash in #2521, and this naming regression, both got past CI. A test app split across a contracts project and a handlers project would close that.

PR #2526, opened 2026-08-12
Generate from an application written as several projects

woksin:
# Summary

Internal end-to-end check and test applications only — no user-facing change.

### DIFF
--- a/.github/workflows/dotnet-build.yml
+++ b/.github/workflows/dotnet-build.yml
@@ -237,6 +237,7 @@ jobs:
         run: |
           dotnet restore TestApps/AspNetCore/AspNetCore.csproj
           dotnet restore TestApps/Chronicle/Chronicle.csproj
+          dotnet restore TestApps/Layered/Layered.csproj
 
       - name: Generate and read back the document of a real application
         run: |
@@ -252,6 +253,19 @@ jobs:
           dotnet Source/DotNET/Screenplay.EndToEnd/bin/Release/net10.0/Cratis.Arc.Screenplay.EndToEnd.dll \
             TestApps/Chronicle/Chronicle.csproj \
             "${RUNNER_TEMP}/Chronicle.play"
+
+      # The two applications above are each one project, and a whole class of defect only exists between projects: a
+      # referenced project arrives as the compilation the workspace built rather than as an assembly on disk, so its
+      # declarations have real source belonging to a compilation that is not the one reading them. Two defects
+      # shipped through that gap and neither was reachable from here until this application existed. It is held to
+      # what its document says as well as to reading back clean, because a generator that quietly declines to read
+      # something writes a smaller document that compiles just as well.
+      - name: Generate and read back the document of a real application written as several projects
+        run: |
+          dotnet Source/DotNET/Screenplay.EndToEnd/bin/Release/net10.0/Cratis.Arc.Screenplay.EndToEnd.dll \
+            TestApps/Layered/Layered.csproj \
+            "${RUNNER_TEMP}/Layered.play" \
+            TestApps/Layered/Layered.expected
 
   proxy-specs:
     needs: [dotnet-build]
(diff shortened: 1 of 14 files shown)


woksin:
## Reviewer context

`TestApps/AspNetCore` and `TestApps/Chronicle` are each **one application project**. A whole class of defect only exists *between* projects: `MSBuildWorkspace` hands a project reference over as the compilation it built rather than as an assembly on disk, so the referenced project's declarations have real source belonging to a compilation that is not the one reading them. Two defects shipped through that gap this week and neither was reachable from CI.

`TestApps/Layered` + `TestApps/Layered.Contracts` is the smallest application that holds both shapes.

### It fails against both pre-fix commits

**Shape 1 — a scenario naming a value the project below it declares** (#2521). The scenario states its customer as `KnownCustomers.House`, a `static readonly CustomerId` declared in `Layered.Contracts`.

Against `ee62638e~1` (`5448a4e9`, before #2522):

```
  project: Layered
Unhandled exception. System.ArgumentException: SyntaxTree is not part of the compilation (Parameter 'syntaxTree')
   at Cratis.Arc.Screenplay.Analysis.Specifications.GeneratedIdentities.<>c__DisplayClass9_0.<Holds>b__1(...)
      in .../Analysis/Specifications/GeneratedIdentities.cs:line 117
   at Cratis.Arc.Screenplay.Analysis.Specifications.SpecificationValues.Add(...)
   at Cratis.Arc.Screenplay.Analysis.Specifications.SpecificationStepReader.ReadWhen(...)
```

The exact production stack, at the exact line.

**Shape 2 — two slices building one read model** (the `SP0034` regression). The placing slice builds `Order` model-bound; the reporting slice builds it again fluently.

Against `205b8903~1` (`28e53268`, before #2523):

```
  SP0034 x1
  generation error SP0034: The generated document did not compile - 1 error(s), the first being
  'Read model 'Order' is built more than once - a projection or a reducer builds it, and only one of them' on line 43
The generated document did not read back clean - 1 diagnostic(s)
EXIT=1
```

### Why the check needed to grow expectations

Reading a document back through `ScreenplayCompiler` proves it is **valid**, not that it is **true**. For shape 1 that gap is not theoretical: whether the cross-project identity is *recognized* or the reader *declines to look past the project boundary*, the value is **absent from the document either way** — `SpecificationValues.Add` omits a recognized generated identity, and omits an unreadable value too. **The documents are byte-identical.** The only difference is that the second reports `SP0040`.

So a document-only assertion would have been vacuous against exactly the trap this fixture exists to catch. `TestApps/Layered/Layered.expected` therefore holds the generation to what was **reported** as well as to what the document says, and the load-bearing line is:

```
reports SP0040 0
```

The directives are `says` (at least once), `once`, `never`, and `reports <code> <n>`. An expectations file is optional, so the other two applications are unaffected. A line that cannot be parsed **throws** rather than being skipped — a file with a typo in it would otherwise pass everything, which is the one outcome worse than failing.

I checked the checker rather than assuming it: deliberately wrong expectations produce

```
  expectation: the document was to say 'something the document never says' at least once, and says it 0 time(s)
  expectation: 'SP0040' was to be reported 3 time(s), and was reported 0 time(s)
  expectation: the document was to say 'concept CustomerId : Uuid' 0 time(s), and says it 1 time(s)
EXIT=1
```

and a typo'd directive throws `UnreadableExpectation`.

### Why not a golden file

I tried that first. Because the fixture project-references Arc's own framework projects, the generated document picks up `module Cratis / feature Arc / slice StateView Queries` from `Cratis.Arc` — so a whole-document golden would churn on unrelated framework changes. (Users referencing Arc from NuGet get metadata references and see no such leakage; this is an artifact of being inside Arc's own repo.) Targeted expectations are immune to it.

### Verification

- `dotnet build Arc.slnx` — Debug and Release, **43 projects, 0 errors, 0 warnings**.
- `Screenplay.Specs` — 1240 passed, 0 failed.
- All three applications through the **Release** binary with the relative paths CI uses: `Layered EXIT=0`, `AspNetCore EXIT=0`, `Chronicle EXIT=0`.

### On the label

Nothing shipped changes — `TestApps/*` and `Screenplay.EndToEnd` are both `IsPackable=false`, and the rest is the workflow. `patch` follows #2455 (TestApp-only, `patch`, "no user-facing change"). Drop it if you would rather not cut a release for it.

Issue #2505, opened 2026-08-10
Arc 21.1.0 cannot start a Development app that uses a Chronicle command response value: its own handler is resolved from the root provider but needs a scoped IEventLog


woksin:
From Arc 21.1.0 onwards, a Development application throws on every command:

```
Cannot resolve 'Cratis.Arc.Chronicle.Commands.SingleEventForEventSourceIdCommandResponseValueHandler'
from root provider because it requires scoped service 'Cratis.Chronicle.EventSequences.IEventLog'.
```

The type is Arc's own. It is reached from the root provider while depending on a scoped service, which is a lifetime mismatch in Arc's registration rather than in the consuming application.

## Why it appears at 21.1.0 and not before

21.1.0 restored the host's scope validation, which Arc had previously been discarding. From its release notes:

> **Changed** — Preserve configured container factories and Development scope validation across the `AddCratisArc` host entry points while disabling only eager validation that Arc cannot support. (#2492)
>
> **Fixed** — Allow `Arc.Core` applications to start in Development without discarding the host's other service-provider settings. (#2492)

That change is correct. It is not the cause of the defect — it is what stops the defect being silently tolerated. The mismatch was presumably always there, and scope validation being switched off is what hid it.

## Isolation

Measured by holding one side fixed and moving the other:

| Arc | Chronicle | Running app |
|---|---|---|
| 21.0.4 | 16.20.0 | starts, commands succeed |
| **21.1.0** | 16.20.0 | **throws on every command** |

So it tracks the Arc version, not the Chronicle version.

## It is invisible to both the compiler and the spec suite

The application compiles clean, and **all 2,029 specs pass either way** — `CommandScenario` builds its own service provider rather than the host's, so no spec ever exercises the root-provider path that fails. The only way to see it is to start the app.

This is the same shape as several other failures found in this wave: a package pair that builds green, specs green, and fails the first time real code runs.

## Consequence for consumers

Cratis Studio is pinned below this. `Cratis.Chronicle.XUnit.Integration` 16.21.0 requires `Cratis.Arc >= 21.1.0`, so taking Chronicle 16.21.0 silently drags an Arc bump and brings this with it. Studio therefore cannot move to Chronicle 16.21.0 or later until this is resolved — independently of [Cratis/cli#67](https://github.com/Cratis/cli/issues/67), which blocks the same upgrade for a different reason (Chronicle 16.20.0's TLS default).

Two unrelated landmines currently gate the same version step.

[...]

PR #2529, opened 2026-08-12
Resolve discovered command and query extension points from the request scope

woksin:
## Fixed

- Start a Development application that uses a Chronicle command response value. Arc's discovered extension points were resolved from the root service provider, so a handler depending on the scoped `IEventLog` could not be created once the host turned scope validation on, and every such command threw. (#2505)
- Resolve command filters, command execution scopes and query filters from the scope the work runs in, so an implementation of any of these that depends on a scoped service works instead of throwing in Development. (#2505)

### DIFF
--- a/Source/DotNET/Arc.Core.Specs/Commands/for_CommandFilters/when_executing_filters/and_the_command_carries_its_own_scope.cs
+++ b/Source/DotNET/Arc.Core.Specs/Commands/for_CommandFilters/when_executing_filters/and_the_command_carries_its_own_scope.cs
@@ -1,0 +1,45 @@
+// Copyright (c) Cratis. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Cratis.Execution;
+using Cratis.Traces;
+
+namespace Cratis.Arc.Commands.for_CommandFilters.when_executing_filters;
+
+/// <summary>
+/// A filter is resolved from the scope the command runs in, not from the provider that constructed this singleton,
+/// so a filter depending on a scoped service is created in the scope rather than in the root — where the container
+/// refuses to create it once scope validation is on.
+/// </summary>
+public class and_the_command_carries_its_own_scope : Specification
+{
+    CommandFilters _commandFilters;
+    ICommandFilter _filterFromTheScope;
+    ICommandFilter _filterFromTheRoot;
+    CommandContext _context;
+    System.Diagnostics.ActivitySource _activitySource;
+
+    void Establish()
+    {
+        _filterFromTheScope = Substitute.For<ICommandFilter>();
+        _filterFromTheRoot = Substitute.For<ICommandFilter>();
+
+        var scope = Substitute.For<IServiceProvider>();
+        scope.GetService(typeof(IInstancesOf<ICommandFilter>))
+            .Returns(new KnownInstancesOf<ICommandFilter>([_filterFromTheScope]));
+
+        _context = new CommandContext(CorrelationId.New(), typeof(object), new object(), [], new(), ServiceProvider: scope);
+
+        var activitySource = Substitute.For<IActivitySource<CommandFilters>>();
+        _activitySource = new System.Diagnostics.ActivitySource("Cratis.Arc.Test");
+        activitySource.ActualSource.Returns(_activitySource);
+        _commandFilters = new(new KnownInstancesOf<ICommandFilter>([_filterFromTheRoot]), activitySource);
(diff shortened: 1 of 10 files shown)


woksin:
Reviewer context.

**The mechanism is one step further down than #2505 describes.** `CommandResponseValueHandlers` does *not* capture the handler instances at construction — `InstancesOf<T>.GetEnumerator()` calls `_serviceProvider.GetService(type)` on **every** enumeration. What it captures is the **`IServiceProvider`**. `IInstancesOf<>` is registered **transient** in Fundamentals (`TypesServiceCollectionExtensions.cs:36`), so it binds to whichever provider resolves it — and a singleton resolves it from the root. The failing stack confirms it:

```
System.InvalidOperationException: Cannot resolve '…HandlerNeedingAScopedCollaborator' from root provider
  because it requires scoped service '…ScopedCollaborator'.
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteValidator.ValidateResolution(…)
   at Cratis.Types.InstancesOf`1.GetEnumerator()+MoveNext()
   at Cratis.Arc.Commands.CommandResponseValueHandlers.CanHandle(…)
```

This matters for the choice of fix: it is not about *when* handlers are built, it is about *which provider* builds them.

**Why resolve-per-invocation rather than a scoped registration.** Making the collections scoped only moves the failure. `CommandPipeline` is a singleton — necessarily, it owns the `IServiceScopeFactory` that creates the scope a command runs in — and it takes `ICommandFilters`, `ICommandResponseValueHandlers` and `ICommandHandlerProviders` as constructor dependencies. A singleton consuming a scoped service is itself a captive dependency, so the container would reject that too. (`CommandResponseValueHandlers` is also registered twice — the explicit `AddSingleton` *and* the `[Singleton]` convention binding — so a lifetime change means changing both.)

Resolving from the scope at the point of use is what Arc already does everywhere else in this pipeline: `TransactionalCommandScope`, `CommandProvideInvoker` and `CommandHandlerArgumentResolver` are all `[Singleton]` and take the scoped provider per invocation, and `IdentityEndpointMapper.cs:134` already resolves `IInstancesOf<T>` from `RequestServices`. The provider is the one `CommandContext`/`QueryContext` already carry, and the HTTP path supplies `context.RequestServices`. It falls back to the injected collection when there is no provider, which is why existing specs are untouched.

**Audit of the siblings in `CommandServiceCollectionExtensions`,** all confirmed by running them against a real Development host rather than by reading:

| Collection | Verdict |
|---|---|
| `ICommandResponseValueHandlers` → `ICommandResponseValueHandler` | **reproduced, fixed** |
| `ICommandFilters` → `ICommandFilter` | **reproduced, fixed** — public extension point |
| `CommandPipeline` → `ICommandExecutionScope` | same shape, fixed — public extension point |
| `ICommandHandlerProviders` → `ICommandHandlerProvider` | **reproduced, deliberately not fixed** — see below |
| `ICommandProvideInvoker`, `ICommandHandlerArgumentResolver` | already correct; take the scoped provider per call |

`ICommandHandlerProvider` fails the same way, but it is not the same defect. It builds the `Type → ICommandHandler` map once at startup, before any scope exists, and that map outlives every scope. A provider requiring a scoped service is a lifetime error in the implementation, not in Arc's registration — there is no correct scope to resolve it from. Left as is, deliberately.

**Scope note — this PR also fixes `QueryFilters`, which is outside the issue.** It is the identical shape (singleton holding `IInstancesOf<IQueryFilter>`, `QueryContext` already carrying the provider), and shipping the command-side fix alone would leave a Development app still throwing for a query filter with a scoped dependency. Say the word and I will split it out.

**Verification.** The four unit specs were run against the unfixed source: **9 of the 10 facts fail**, and the host spec fails with the exception above. With the fix, all pass. Debug and Release builds clean with zero warnings on `Arc.Core`, `Arc`, `Chronicle` and the spec projects. Specs: `Arc.Core.Specs` 1925, `Chronicle.Specs` 632, `Arc.Specs` 274, `MongoDB.Specs` 192, `Swagger.Specs` 47, `Testing.Specs` 19, `OpenApi.Specs` 8 — **3,097 passing, 0 failing**. The full-solution build could not run in this worktree because `ProxyGenerator.Specs` shells out to `yarn build` and the worktree has no `node_modules`; that is a worktree artifact and unrelated to the change — CI covers it.

**Not verified:** no Chronicle-backed application was started end to end against these bits. The host spec reproduces the exact container failure with a stand-in scoped service rather than the real `IEventLog`.

Issue #2478, opened 2026-08-10
Add a production-named fixed tenancy registration API


woksin:
## Problem

Arc 21.0.4 exposes `UseDevelopmentTenancy(tenantId)` as the only fixed-tenant registration API, but its behavior is not development-only. It registers a fixed resolver in every build configuration. The resulting `ITenantIdAccessor` also drives Chronicle namespace and Mongo database resolution, including request-free hosted `ICommandPipeline` work.

A production service with one tenant therefore has to call an API named "Development" or build a custom resolver solely to avoid the misleading name.

## Verified boundary

At source-link commit `dc7e9e362969fd2139b119e67e9bff3167aeb524` behind Arc 21.0.4:

- `UseDevelopmentTenancy()` selects `DevelopmentTenantIdResolver` without checking the host environment.
- `TenantIdAccessor` exposes that fixed value.
- Arc Chronicle's `TenantNamespaceResolver` and Arc MongoDB's `DefaultMongoDatabaseNameResolver` consume the same accessor.
- `ICommandPipeline` has no tenant argument and `ITenantIdAccessor` has no ambient setter/scope.

This was found while implementing a one-shot hosted bootstrap command for a fixed-tenant production application. The correct native mechanism exists; only its public name says it is inappropriate.

## Proposed compatible shape

Add an additive `UseFixedTenancy(tenantId)` registration API and production-neutral fixed-resolver naming. It may delegate to the existing implementation. Preserve `UseDevelopmentTenancy` as a compatibility alias (with a normal deprecation window if desired).

This does **not** ask Arc to adopt any application's tenant id or single-tenant policy. A future ambient tenant-execution scope for multi-tenant background jobs is a separate design.

## Acceptance

- `UseFixedTenancy("tenant-a")` resolves `tenant-a` outside Development.
- Chronicle namespace and Mongo database resolution consume the same accessor.
- Existing `UseDevelopmentTenancy` behavior remains unchanged.
- Documentation distinguishes fixed tenancy from request-resolved tenancy.

This is additive and fits a minor release.

woksin:
Accepted as a real defect on Arc's own terms, and being implemented.

[...]

PR #2539, opened 2026-08-12
Add a production-named fixed tenancy registration API

woksin:
## Added

- `UseFixedTenancy` and `FixedTenantIdResolver`, giving single-tenant production deployments a registration API that is not named for the development environment (#2478)
- `Tenancy:FixedTenantId` configuration, which `Tenancy:DevelopmentTenantId` now forwards to (#2478)

## Fixed

- The unknown tenant-resolver-type error now lists `Subdomain` and `Fixed`, which it omitted (#2478)

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -13,6 +13,10 @@
     <!-- Microsoft.OpenApi 2.0.0 (pulled in transitively by Microsoft.AspNetCore.OpenApi) has a high-severity
          advisory (GHSA-v5pm-xwqc-g5wc). Pin to the patched 2.x release. -->
     <PackageVersion Include="Microsoft.OpenApi" Version="2.7.5" />
+    <!-- SSH.NET 2025.1.0 (pulled in transitively by Testcontainers) has a high-severity advisory
+         (GHSA-q939-rpr3-3284). Testcontainers 4.13.0 is the latest and still resolves the vulnerable
+         version, so pin to the patched release. -->
+    <PackageVersion Include="SSH.NET" Version="2026.0.0" />
     <!-- Cratis -->
     <PackageVersion Include="Cratis.Chronicle" Version="16.26.0" />
     <PackageVersion Include="Cratis.Chronicle.AspNetCore" Version="16.26.0" />
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -13,10 +13,6 @@
     <!-- Microsoft.OpenApi 2.0.0 (pulled in transitively by Microsoft.AspNetCore.OpenApi) has a high-severity
          advisory (GHSA-v5pm-xwqc-g5wc). Pin to the patched 2.x release. -->
     <PackageVersion Include="Microsoft.OpenApi" Version="2.7.5" />
-    <!-- SSH.NET 2025.1.0 (pulled in transitively by Testcontainers) has a high-severity advisory
-         (GHSA-q939-rpr3-3284). Testcontainers 4.13.0 is the latest and still resolves the vulnerable
-         version, so pin to the patched release. -->
-    <PackageVersion Include="SSH.NET" Version="2026.0.0" />
     <!-- Cratis -->
     <PackageVersion Include="Cratis.Chronicle" Version="16.26.0" />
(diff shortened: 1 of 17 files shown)


woksin:

**Premise verified on the branch, not from the issue.** `DevelopmentTenantIdResolver.Resolve()` was `=> options.Value.Tenancy.DevelopmentTenantId;` with no `IHostEnvironment` reference anywhere under `Tenancy/`, and `HostBuilderExtensions.cs:103` selected it unconditionally from `ResolverType`.

**Backward compatibility.** Purely additive: `Fixed = 5` is the next free enum value with 0–4 untouched; `DevelopmentTenantIdResolver` keeps its public name and constructor signature and now derives from `FixedTenantIdResolver`; `UseDevelopmentTenancy` is byte-identical and deliberately not deprecated. No existing public method gained a parameter — the C# hazard is that a trailing optional parameter is binary breaking because the default is baked into the call site, so `UseFixedTenancy` being a new method is what makes its optional parameter safe.

**Configuration binding, checked rather than assumed.** Binding is plain reflection `ConfigurationBinder` via `Configure<ArcOptions>(...)`, which writes through public setters, so the forwarding property works and both `Tenancy:FixedTenantId` and `Tenancy:DevelopmentTenantId` land on the same value. Two specs bind a real `ConfigurationBuilder` and assert each key under both names. Caveat now documented in `configuration.md` and the XML docs: if a source supplies **both** keys, whichever the binder visits last wins, so supply only one.

**Mutation evidence.** Each apply→run→restore in one command with `trap … EXIT`, original bytes restored, verified by `shasum -a 256` — all three matched.
- Registration arm returns `DevelopmentTenantIdResolver` instead of `FixedTenantIdResolver` → `and_fixed_tenancy_is_configured` RED on the exact-type assertion. **SPECIFIC.**
- `FixedTenantIdResolver.Resolve()` → `"MUTANT"` → `for_FixedTenantIdResolver` RED 3/3, **SPECIFIC**.
- `DevelopmentTenantId` setter made a no-op → the configuration-forwarding spec RED. **SPECIFIC.**

**Two findings reported rather than papered over.**
1. *A surviving mutation.* The registration mutation's second assertion, `should_resolve_the_configured_tenant_id`, survives — with the arm pointing at `DevelopmentTenantIdResolver`, `Resolve()` still returns the right value because both types read `FixedTenantId`. Only the exact-type assertion distinguishes the two registrations, so `ShouldBeOfExactType<FixedTenantIdResolver>` is load-bearing; a `ShouldBeOfType`/`is` assertion there would be vacuous, since `DevelopmentTenantIdResolver` *is* a `FixedTenantIdResolver`.
2. *Killed by the wrong spec.* Because `DevelopmentTenantIdResolver` no longer declares `Resolve()`, a mutation in `FixedTenantIdResolver` reddens the pre-existing `for_DevelopmentTenantIdResolver` specs **collaterally**. Those specs still pass and still have value — they pin the `DevelopmentTenantId` forwarding, which the setter mutation kills specifically — but they no longer pin anything unique to the derived type. That is inherent to one-implementation delegation, not something the specs can fix.

**Gate.** Debug `--no-incremental`, `Arc.Core.Specs` (builds `Arc.Core` plus both analyzer/generator projects): 0 errors, 0 warnings; 1936 passed / 0 failed, re-run after both commits. Release `--no-incremental`: 0 errors, 0 warnings for each of `Arc.Core`, `Arc.Core.Specs`, `Arc`, `Arc.Specs`, `Chronicle`, `MongoDB`, `Testing`, `OpenApi`, `Swagger` and `TestApps/ArcCore`. Zero generated-proxy churn — nothing under `Source/JavaScript` changed.

A full `Arc.slnx -c Release` build fails on `ProxyGenerator.Specs` with `MSB3073: "yarn build" exited with code 1`, which is environmental: that target shells out to `yarn build` and this worktree has no `node_modules`. No JavaScript was touched and `yarn install` was deliberately not run here.

**Not verified:** no runtime check that `Fixed` flows through to Chronicle namespace or Arc Mongo database resolution — only that the correct `ITenantIdResolver` is selected from DI. Other Cratis repos were not searched for consumers of `DevelopmentTenantIdResolver` that the new base class could affect. The two changed docs pages were not rendered.

PR #2540, opened 2026-08-13
Declare server-consumed command response types

### LINKED ISSUE #2537: Let a command response value handler declare its consumed type so the proxy generator stops emitting it
`Source/DotNET/Tools/ProxyGenerator/TypeExtensions.cs:91-94` hardcodes the two Arc.Core-handled types as the generator's only notion of "consumed by a handler, not returned to the client":

    static readonly HashSet<string> _wellKnownTypeNames =
    [
        "Cratis.Arc.Validation.ValidationResult",
        "Cratis.Arc.Authorization.AuthorizationResult"
    ];

Arc ships **eight** response-value handlers. The six in `Source/DotNET/Chronicle/Commands/` — `EventsCommandResponseValueHandler`, `EventsWithConcurrencyScopesCommandResponseValueHandler`, `AggregateRootCommitResultCommandResponseValueHandler`, `SingleEvent…` and the two `…ForEventSourceId` variants — are invisible to the generator.

The consequence: a command returning `Result<EventsWithConcurrencyScopes, ValidationResult>` generates as `Command<IX, EventsWithConcurrencyScopes>` and recursively emits proxies for `EventsWithConcurrencyScopes`, `ConcurrencyScope` and `KeyValuePair<…>` — none of which ever cross the wire, because the handler consumes the value and returns a plain successful `CommandResult`.

Runtime cannot help the generator today: `ICommandResponseValueHandler.cs:17` is `bool CanHandle(CommandContext commandContext, object value)` — instance-based, with no type-level declaration a generator could read.

**Proposal.** Add an additive, generator-visible contract — e.g. `ICanDeclareConsumedResponseValueTypes { IEnumerable<Type> ConsumedTypes { get; } }` — implemented by every built-in handler, and have the generator read it through the metadata load context instead of the literal list. Runtime and generator then share one source of truth, and a custom handler gets the same behavior for free.

**Acceptance:**
- A model-bound command returning a handler-consumed type generates as a no-response command.
- No proxy is emitted for that type or its property graph solely because of the return.
- An ordinary unhandled response type is still emitted normally.
- The two Arc.Core types keep their current behavior via the new contract, and the literal list is deleted.

Scoped down from #1493, which bundled this with an open-ended "revision of the whole response value system". The filtering half of that issue was separately addressed by the build-level `<ExcludeType>` surface (`54fa9e2d`, 2026-05-03) and type mappings (`b89e9ef1`).

Semver: minor — a new optional interface; `ICommandResponseValueHandler` is unchanged.

woksin:
# Summary

Command response value handlers can now declare the response type they consume, and the proxy generator uses these declarations to stop generating client response models for values that are handled on the server and never reach the client.

## Added

- `ICommandResponseValueHandler<TValue>`, an optional typed declaration that marks a command response value type as consumed on the server (#2537)

## Changed

- The proxy generator no longer generates client response models for server-consumed response types, whether returned directly or wrapped in `Result`, `OneOf`, tuples, or collections (#2537)
- Typed handler declarations are discovered from the application and its referenced packages, replacing the fixed built-in list of well-known types (#2537)
- All built-in Arc and Chronicle response value handlers declare their consumed types (#2537)

## Fixed

- Generated TypeScript import statements for generic types now reference the correct proxy file name

### DIFF
--- a/Arc.slnx
+++ b/Arc.slnx
@@ -39,6 +39,8 @@
         <Project Path="Source\DotNET\Tools\ProxyGenerator\ProxyGenerator.csproj" />
         <Project Path="Source\DotNET\Tools\ProxyGenerator.Build\ProxyGenerator.Build.csproj" />
         <Project Path="Source\DotNET\Tools\ProxyGenerator.Specs\ProxyGenerator.Specs.csproj" />
+        <Project Path="Source\DotNET\Tools\ProxyGenerator.Specs\TestArtifacts\CommandResponseHandlerDependency\CommandResponseHandlerDependency.csproj" />
+        <Project Path="Source\DotNET\Tools\ProxyGenerator.Specs\TestArtifacts\FakeCommandResponseHandlerDependency\FakeCommandResponseHandlerDependency.csproj" />
     </Folder>
     <Folder Name="/TestApps/">
         <Project Path="TestApps\AspNetCore\AspNetCore.csproj" />
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -7,6 +7,9 @@
   </PropertyGroup>
   <ItemGroup>
     <!-- Transitive security pins -->
+    <!-- SSH.NET through Testcontainers <= 2025.1.0 has a high-severity advisory
+         (GHSA-q939-rpr3-3284). Pin the transitive dependency to the patched release. -->
+    <PackageVersion Include="SSH.NET" Version="2026.0.0" />
     <!-- SQLitePCLRaw 2.1.10/2.1.11 (pulled in transitively by EF Core Sqlite) has a high-severity advisory
          (GHSA-2m69-gcr7-jv3q). Pin the native library to the patched release. -->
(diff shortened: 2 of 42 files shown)


woksin:
Reviewer context moved out of the description (release notes are generated from it):

**Internal changes** (no user-facing effect): metadata load-context ownership is bounded to each generator invocation; dynamically-registered handlers stay runtime-only so ordinary DTO responses are not suppressed; SSH.NET pinned past GHSA-q939-rpr3-3284 (transitive test-only dependency via Testcontainers); proxy generator test artifacts added to the solution graph.

**Verification** (from the original description):
- Debug ProxyGenerator.Specs build: 0 warnings, 0 errors
- Focused response/discovery/lifecycle filter: 37/37
- Full Debug ProxyGenerator.Specs: 1141/1141 before the final counterfeit-contract negative
- Release build (net8/net9/net10): 0 warnings, 0 errors
- Final full Release net10 ProxyGenerator.Specs: 1143/1143
- Independent code/compatibility and security reviews: approved
- `git diff --check`: clean

